PR #1, opened 2025-10-14
chore: sets the initial config for a market place and a hello world plugin

enxebre:
This enable consumers to install reusable pieces from this repo https://www.anthropic.com/news/claude-code-plugins https://anthropic.mintlify.app/en/docs/claude-code/plugin-marketplaces#how-to-configure-team-marketplaces https://anthropic.mintlify.app/en/docs/claude-code/plugins https://claudecodemarketplace.com/marketplace/anthropic-claude-code

### DIFF
--- a/plugins/hello-world/commands/echo.md
+++ b/plugins/hello-world/commands/echo.md
@@ -1,0 +1,75 @@
+---
+description: Hello world plugin implementation
+argument-hint: [name]
+---
+
+## Name
+hello-world:echo
+
+## Synopsis
+```
+/hello-world:echo [name]
+```
+
+## Description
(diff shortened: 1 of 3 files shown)


plugins/hello-world/commands/echo.md
```diff
@@ -0,0 +1,7 @@
+---
+description: Hello world plugin implementation
+---
+
+# Hello Command
```
brandisher: In relation to [this comment](https://github.com/openshift-eng/ai-helpers/pull/2/files#r2428827519), what do you think about the this command being the definition for what new commands should look like?

I understand the purpose is to seed the marketplace but we could accomplish two goals with one PR.

enxebre: I really like this idea. I updated it to serve as a ref implementation

PR #2, opened 2025-10-14
feat(plugin): introduce jira plugin with solve command

enxebre:
This brings the HCP jira-solve command from https://github.com/openshift/hypershift/blob/main/.claude/commands/jira-solve.md.
Intent is to abstract it away and let it be used across any repo based on certain conventions

Goes after https://github.com/openshift-eng/ai-helpers/pull/1

### DIFF
--- a/plugins/jira/.claude-plugin/plugin.json
+++ b/plugins/jira/.claude-plugin/plugin.json
@@ -1,0 +1,8 @@
+{
+  "name": "jira",
+  "description": "A plugin to automate tasks with Jira",
+  "version": "0.0.1",
+  "author": {
+    "name": "github.com/openshift-eng"
+  }
+}
--- a/plugins/jira/commands/solve.md
+++ b/plugins/jira/commands/solve.md
@@ -1,0 +1,114 @@
+---
+description: Analyze a JIRA issue and create a pull request to solve it.
+---
+
+Analyze a JIRA issue and create a pull request to solve it.
+
+[Extended thinking: This command takes a JIRA URL, fetches the issue description and requirements, analyzes the codebase to understand how to implement the solution, and creates a comprehensive pull request with the necessary changes.]
+
+**JIRA Issue Analysis and PR Creation**
+
+## Usage Examples:
+
+1. **Solve a specific JIRA issue**:
+   `/jira-solve OCPBUGS-12345 enxebre`
+
+## Implementation Details:
+
+- The command uses curl to fetch JIRA data via REST API: https://issues.redhat.com/rest/api/2/issue/{$1}
+- Parses JSON response using jq or text processing
+- Extracts key fields: summary, description, components, labels
+- No authentication required for public Red Hat JIRA issues
+- Creates a PR with the solution
+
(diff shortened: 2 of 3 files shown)


plugins/jira/commands/solve.md
```diff
@@ -0,0 +1,114 @@
...
+   - After creating the PR, display the PR URL and description to the user
+   - Ask the user: "Please review the PR description. Would you like me to update it? (yes/no)"
+   - If the user says yes or requests changes:
+     - Ask what changes they'd like to make
+     - Update the PR description using `gh pr edit {PR_NUMBER} --body "{new_description}"`
+     - Repeat this review step until the user is satisfied
+   - If the user says no or is satisfied, acknowledge and provide next steps
+
+
+## Arguments:
```
brandisher, line 110: If I've understood correctly, this is a port of a slash command from another source so take this as a nit. It'd make sense to standardize on known command conventions. Here's an example:

```
ls --help
Usage: ls [OPTION]... [FILE]...
List information about the FILEs (the current directory by default).
Sort entries alphabetically if none of -cftuvSUX nor --sort is specified.
```

Translating this to slash commands the format would be
```
1. Usage
2. Description
3. Arguments
4. Everything else
```
This format could be easily interrogated on PRs, and gives users an expectation on how to find the info that they need in the command.

enxebre: Makes sense. I updated https://github.com/openshift-eng/ai-helpers/pull/1 to serve as the ref implementation.
This one has been under usage for a little while in the hypershift repo and we've been developing empirical knowledge for when it works and when it doesn't work well. I'd rather merge it as it's so folks can start using it right away without risk for regressing and then iterate transparently to match conventions set by https://github.com/openshift-eng/ai-helpers/pull/1

plugins/jira/commands/solve.md
```diff
@@ -0,0 +1,114 @@
...
+   - Understand current implementation
+   - Identify areas that need changes
+   - Use Grep and Glob tools to search for:
+      - Related function names mentioned in JIRA
+      - File patterns related to the component
+      - Similar existing implementations
+      - Test files that need updates
+
+3. **Solution Implementation**:
+   - Think hard and create a detailed, step-by-step plan to implement this feature. Save it to spec-$1.md within the contrib/ folder, for example contrib/spec-OCPBUGS-12345.md
```
mrunalp, line 50: Should we invoke  "ultrathink" here?

plugins/jira/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "jira",
+  "description": "A plugin to automate tasks with Jira",
+  "version": "0.0.1",
+  "author": {
+    "name": "github.com/enxebre"
```
bryan-cox: Would we instead put the hypershift repo here? 🤔

enxebre: actually I think https://github.com/openshift-eng or an alias or an individual would be a better contact for this

PR #13, opened 2025-10-16
feat(plugin): Add yaml-docs slash command

saswatamcode:
Something I've been using as a slash command for some time.

yaml-docs allows me to point to any Go struct and generate a markdown code block with YAML of the same struct.
Particularly useful when iterating on configuration docs of any kind.

There are deterministic ways of doing this in Go with reflect and structtags, but those require more effort. 

This generates decent, fully-qualified YAML, with all fields populated with best guess values, that a human author can modify once generated.

This PR also adds a marketplace, for these plugins.

### DIFF
--- a/plugins/yaml/commands/docs.md
+++ b/plugins/yaml/commands/docs.md
@@ -3,9 +3,31 @@ description: Generate comprehensive YAML documentation from Go struct definition
 argument-hint: [file:StructName] [output.md]
 ---
 
+## Name
+yaml:docs
+
+## Synopsis
+```
+/yaml:docs [file:StructName] [output.md]
+```
+
+## Description
+The `yaml:docs` command generates comprehensive YAML documentation from Go struct definitions. It analyzes Go structs and produces complete, well-documented YAML configuration examples with intelligent default values for all fields.
+
+This command is designed to help developers quickly create YAML configuration documentation by:
+- Automatically generating sensible default values for all struct fields
+- Adding inline comments explaining each field's purpose and constraints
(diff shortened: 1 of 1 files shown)


plugins/.claude-plugin/marketplace.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "ai-helpers",
+  "owner": {
+    "name": "openshift-eng"
+  },
+  "plugins": [
+    {
+      "name": "yaml-docs",
```
enxebre: this would result in the command being invoked as via `/yaml-docs:yaml-docs` so may be name the plugin yaml and the command docs so then you install the yaml plugin and you can `/yaml:docs`

saswatamcode: Yup converted this from a slash command, so missed that.
Updated!

plugins/yaml/commands/docs.md
```diff
@@ -0,0 +1,120 @@
+---
+description: Generate comprehensive YAML documentation from Go struct definitions with sensible default values
+argument-hint: [file:StructName] [output.md]
+---
+
+You are a specialized tool for generating comprehensive YAML documentation from Go struct definitions.
+
```
enxebre, line 29: can you please adjust to adhere to the ref implementation ## sections https://github.com/openshift-eng/ai-helpers/blob/main/plugins/hello-world/commands/echo.md

saswatamcode: Thanks for the review, have tried to adhere to the format!

PR #24, opened 2025-10-16
Adding instructions on setting up a local jira mcp server using podman

calfonso:
This PR provides instructions to set up a jira mcp server using podman

## Checklist:
- [x] Relevant issues have been referenced.

### DIFF
--- a/plugins/jira/README.md
+++ b/plugins/jira/README.md
@@ -22,7 +22,40 @@ Comprehensive Jira integration for Claude Code, providing AI-powered tools to an
 claude mcp add atlassian npx @modelcontextprotocol/server-atlassian
 ```
 
+OR you can have claude use an already running Jira MCP Server
+
+```bash
+# Add the Atlassian MCP server
+claude mcp add --transport sse atlassian http https://localhost:8080/sse
+```
+
 Configure your Jira credentials according to the [Atlassian MCP documentation](https://github.com/modelcontextprotocol/servers/tree/main/src/atlassian).
+
+### Running Jira MCP Server locally with podman
+
+```bash
+# Start the atlassian mcp server using podman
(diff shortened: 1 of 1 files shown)


plugins/jira/README.md
```diff
@@ -14,6 +14,29 @@ Comprehensive Jira integration for Claude Code, providing AI-powered tools to an
 - Claude Code installed
 - Jira MCP server configured
 
+### Running Jira MCP Server locally
```
bryan-cox: Should the block from L40 to L47 be moved up above this so that the setup for Jira with Claude and the setup with Jira MCP locally are together?

calfonso: Good call out. I actually think it should be structured even more than that after your comment made me think more. Either pick npm install OR podman, THEN add via claud mcp or config file (because you need to know the endpoint when you add to claude.) Thoughts?

plugins/jira/README.md
```diff
@@ -14,6 +14,29 @@ Comprehensive Jira integration for Claude Code, providing AI-powered tools to an
...
+### Getting Tokens 
+You'll need to generate your own tokens in several of these examples:
+
+- For JIRA API TOKEN, use https://id.atlassian.com/manage-profile/security/api-tokens
+- For JIRA PERSONAL TOKEN, use https://issues.redhat.com/secure/ViewProfile.jspa?selectedTab=com.atlassian.pats.pats-plugin:jira-user-personal-access-tokens
+- For GitHub bearer token, use https://github.com/settings/tokens
+
+### Notes and tips
+
+- Do not commit real tokens. If you must keep a project-local file, prefer committing a `mcp.json.sample` with placeholders, and keep your real `mcp.json` untracked.
```
stbenjam, line 50: People should be using the pre-commit hook from IT as well, maybe we could inform them about that here?

PR #68, opened 2025-10-27
Slash command for installing on vsphere

jcpowermac:

Installing OpenShift on vSphere, in our dev environments, can be difficult for a first timer. 
This slash command (with skills, scripts to reduce token usage) flows through the usual
steps that someone needs to perform to install vSphere in our environments

## Checklist:
- [ ] Subject and description added to both, commit and PR.
- [ ] This change includes docs.

### DIFF
--- a/plugins/openshift/README.md
+++ b/plugins/openshift/README.md
@@ -4,6 +4,18 @@ OpenShift development utilities and workflow helpers for Claude Code.
 
 ## Commands
 
+### `/openshift:install-vsphere`
+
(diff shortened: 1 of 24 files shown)


jcpowermac:
> thanks! is there anything specific about destructive / write operations for the CLIs used that needs to be called out explicitly in the spec for the model? /approve

It will download the ova, and then upload if there is not a template already available. I can't think of anything destructive

Looking through what Claude wrote yesterday, it missed alot, even after we went through testing it.

stbenjam:
@jcpowermac This PR is approved, I'd suggest getting anyone with vpshere knowledge to lgtm it for you

PR #134, opened 2025-11-07
Improve git:cherry-pick-by-patch command

mpatlasov:
Let AI to apply changes manually for conflicted files. See commit description for details.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved patch-apply error handling: when a patch fails to apply cleanly, the workflow now detects conflicted files and attempts manual conflict resolution.
  * Filenames needing manual changes are reported for transparency.
  * The operation now only fails if required manual changes cannot be applied, instead of failing immediately on a non-zero patch exit.

### DIFF
--- a/plugins/git/commands/cherry-pick-by-patch.md
+++ b/plugins/git/commands/cherry-pick-by-patch.md
@@ -37,7 +37,9 @@ Fail, if there is no `commit_hash` in the current repository checkout.
     ```bash
     git show commit_hash | patch -p1 --no-backup-if-mismatch
     ```
-and check if exit code is zero. Fail if exit code is not zero.
+and check if exit code is zero. If exit code is not zero, find conflicted files in the output and apply
+changes for them from the patch manually. Print file names which required manual changes. Fail workload
+if you cannot apply any required change manually.
 
 2. Find files removed from local checkout by the patch command and execute `git rm` for them.
 


mpatlasov:
@enxebre , Can you please lgtm/approve this PR. This is a simple follow-up to [PR#50 you reviewed before](https://github.com/openshift-eng/ai-helpers/pull/50#issuecomment-3441867156). This PR [let me automate cherry-pick](https://github.com/openshift/csi-operator/pull/465) where command failed before. So this useful practically.
/assign @enxebre

plugins/git/commands/cherry-pick-by-patch.md
```diff
@@ -37,7 +37,9 @@ Fail, if there is no `commit_hash` in the current repository checkout.
     ```bash
     git show commit_hash | patch -p1 --no-backup-if-mismatch
     ```
-and check if exit code is zero. Fail if exit code is not zero.
+and check if exit code is zero. If exit code is not zero, find confilcted files in the output and apply
```
zaneb:
```suggestion
and check if exit code is zero. If exit code is not zero, find conflicted files in the output and apply
```

mpatlasov:
> I don't understand why this command exists (surely `git cherry-pick` followed by manually fixing the conflicts is strictly better than `git show | patch` and manually fixing the conflicts?)

Thank you, @zaneb ! (this command has a practical use: in many cases it replaces "manually fixing the conflicts" with resolving conflicts by AI)

zaneb (commented):
I meant I'd have thought `git cherry-pick` followed by AI fixing the conflicts is better than `git show | patch` and AI fixing the conflicts, since the latter throws away information that only git has.

mpatlasov:
> So basically this looks like an expensive way of injecting a higher risk of error.

OK, I understand your point now, thank you! For now, I have two practical use-cases when my command (based on "patch") worked perfectly fine. I'm going to continue to use this command for backporting patches wherever ordinary `git cherry-pick` fails. In case it injects an error, of course, I'll try to fix it, may be including switch from "patch" to "git cherry-pick" on the first step.

For now, the command is optional, and rather light-weight (according to `/cost` output). Anyone who uses it, is supposed to verify results by eyes -- reviewing the final commit. This would remain the case even if we switch from the "patch" to "git cherry-pick". Generally, I'm rather neutral about patch-vs-git-cherry-pick controversial. As soon as we gather more data confirming that one approach is more reliable than another, I'm OK to rework this command, but such decision must be justified.

I appreciate your involvement, @zaneb ! Before this conversation I didn't realize that "git cherry-pick" and "patch" can be used interchangeably, even if both fail due to conflicts. And I didn't realize that "patch" is almost twice as long and was more than twice as expensive. This is a good point to have in mind in future!

Btw, can you give me some details how I can reproduce these numbers:
```
AI resolving conflicts:
        claude-haiku-4-5:  2.1k input, 134 output, 0 cache read, 0 cache write
        claude-sonnet-4-5:  2.3k input, 4.8k output, 161.6k cache read, 32.2k cache write

git:cherry-pick-by-patch command:
        claude-haiku-4-5:  11.8k input, 1.6k output, 0 cache read, 6.0k cache write
        claude-sonnet-4-5:  326 input, 6.4k output, 1.0m cache read, 23.4k cache write
```
I understand that claude-haiku-4-5 vs. claude-sonnet-4-5 are different AI models used by Claude. But which command text or environment differentiates your "AI resolving conflicts" vs. "git:cherry-pick-by-patch command"?

PR #177, opened 2025-11-27
Add ci:pr-retest plugin command

jluhrsen:
will look at a PR and give options for retesting e2e's and any payload jobs that may have failures

I'm not sure how much "AI" this PR is currently, but it is a starting point. There are two bash scripts that
analyze CI jobs (e2e and payload jobs) for failures and quickly gives a summary and options for retesting
if the user wants.

You can run it as a /slash command from claude which adds all the AI overhead and actually takes
significantly longer to run the two bash scripts.

adding more understanding to the /ci:pr-retest slash command could expand this to help understand
if any of the job failures have typical patterns worth a deeper investigation or just "one more" /retest
would be ok.

## Summary by CodeRabbit

* **New Features**
  * Added /ci:e2e-retest, /ci:payload-retest, and /ci:pr-retest commands to locate and retest failed e2e jobs, payload jobs, or both on a PR.
  * Commands support repo auto-detection and repo argument variants, interactive retest options (selective, all, or /retest), and a combined PR-level wrapper.

* **Documentation**
  * Added user-facing docs with usage, examples, workflows, and expected outputs for the new retest commands.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -26,7 +26,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-      "version": "0.0.7"
+      "version": "0.0.8"
     },
     {
       "name": "teams",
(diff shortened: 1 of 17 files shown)


bentito:
Well I definitely get the idea of a base tool in a toolset but I think you should do at least a bit of work that's only possible with an LLM involved in each. You can definitely build in this as a base by referencing it in your later tools

PR #277, opened 2026-01-08
Add /openshift:sniff-test command to review code for quality issues

dobsonj:
This PR introduces a /sniff-test command for claude code to review a file or commit for "code smells", or bad practices that should be addressed. This can be used to find new issues in existing code, or it can be used on your local commits to prevent introducing new issues in a PR.

I originally introduced this for local-storage-operator in https://github.com/openshift/local-storage-operator/pull/571 and it turns out to be fairly useful. @mpatlasov added it to https://github.com/openshift/csi-operator/pull/486 as well. We want to consolidate it here since it seems to be generally useful.

## Summary by CodeRabbit

* **New Features**
  * Added `/openshift:sniff-test <file|commit>` command to the OpenShift plugin. Users can now analyze code quality, detect bad patterns, and identify code smells across specified files or commits.

* **Documentation**
  * Updated plugin documentation with details on the new sniff-test command and its usage.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/plugins/openshift/commands/sniff-test.md
+++ b/plugins/openshift/commands/sniff-test.md
@@ -1,0 +1,66 @@
+---
+argument-hint: <file|commit>
+description: Review code for quality issues, bad patterns, and code smells.
+---
+
+## Name
+openshift:sniff-test
+
+## Synopsis
+```
+/openshift:sniff-test <file|commit>
+```
+
+## Description
+
+Review code for quality issues, bad patterns, and code smells.
+
+This command accepts a file path or commit hash as an argument to analyze for code quality issues.
+Usage: `/openshift:sniff-test path/to/file.go` or `/openshift:sniff-test commit-hash`
+
+## Implementation
+
+The command will analyze the code for:
+- Bad patterns or anti-patterns
+- Unnecessary or dead code
+- Unreferenced functions or variables
+- Code duplication
+- Functions that should be simplified or refactored
+- API fields missing documentation
+- Hardcoded values that should be constants
+- Unhelpful or misleading comments
+- Missing error handling
+- Performance issues
+- Security vulnerabilities
+- Inconsistent naming conventions
+- Missing unit tests for public functions
+
+Steps:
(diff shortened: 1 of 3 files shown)


plugins/openshift/commands/sniff-test.md
```diff
@@ -0,0 +1,66 @@
...
+- Security vulnerabilities
+- Inconsistent naming conventions
+- Missing unit tests for public functions
+
+Steps:
+1. Determine if the argument is a file path or commit hash
+2. If it's a commit hash, get the list of changed files in that commit
+3. For each file to analyze:
+   - Read the file content
+   - Perform static analysis looking for code smells
```
theobarberbany, line 43: Do we want to call out specific static analysis tools here that may be useful (and reduce the non-determinism / token budget spend?). AI is typically not as good as those dedicated tools. 

  The step could become: "Run golangci-lint run (or project-configured linter) on the file(s) and incorporate results, then focus AI analysis on
  design-level concerns that linters don't cover."


  Concrete tools worth calling out for Go:

  - go vet — catches common mistakes (printf format mismatches, unreachable code, etc.)
  - staticcheck — broader set of checks, well-maintained
  - golangci-lint — meta-linter that runs many linters in one pass
  - go vet -vettool with custom analyzers if the project has any

plugins/openshift/commands/sniff-test.md
```diff
@@ -0,0 +1,66 @@
...
+
+This command accepts a file path or commit hash as an argument to analyze for code quality issues.
+Usage: `/openshift:sniff-test path/to/file.go` or `/openshift:sniff-test commit-hash`
+
+## Implementation
+
+The command will analyze the code for:
+- Bad patterns or anti-patterns
+- Unnecessary or dead code
+- Unreferenced functions or variables
```
theobarberbany, line 26: The Go compiler already rejects unused imports and variables. Unused functions is the only
  useful check here, and deadcode or staticcheck with U1000 handles it better.

PR #418, opened 2026-04-14
ci: add parse-junit skill for JUnit XML parsing

stbenjam:
## Summary

- Adds a new `parse-junit` skill to the CI plugin with a stdlib-only Python script that parses JUnit XML files from OpenShift CI jobs
- The payload agent reinvents JUnit XML parsing inline ~500+ times across sessions — this gives it a reusable tool instead
- Handles standard JUnit, aggregated JUnit (with system-out YAML for per-run job URLs), gzip-compressed files, and stdin input
- Extracts test metadata including suite/binary source, lifecycle (informing vs blocking), failure messages, and output text
- Updates `prow-job-analyze-test-failure` to use `parse-junit` instead of inline XML parsing instructions
- Bumps CI plugin version to 0.0.38

## Test plan

- [x] Verified `--help` output
- [x] Tested standard JUnit XML parsing (pass/fail/skip detection, failure messages)
- [x] Tested informing lifecycle auto-detection from suite name
- [x] Tested aggregated JUnit system-out YAML parsing (per-run job URLs)
- [x] Verified all output formats: json, summary, failures, names
- [x] Ran `make update` — docs and marketplace synced
- [x] Ran `make lint` — passes

## Summary by CodeRabbit

* **New Features**
  * Added Parse JUnit XML skill to extract and analyze test results with filtering by test name, status, and lifecycle, supporting multiple output formats (JSON, summary, failures-only).

* **Documentation**
  * Added comprehensive documentation for the new Parse JUnit XML skill with CLI usage and examples.
  * Updated Prow Job test failure analysis documentation to leverage the new Parse JUnit XML skill.

* **Chores**
  * Updated CI plugin version to 0.0.43.


stbenjam:
/hold

Want to test against a few payloads first

plugins/ci/skills/parse-junit/parse_junit.py
```diff
@@ -0,0 +1,556 @@
...
+                status = "failed"
+                failure_message = failure_el.get("message", "")
+                failure_text = failure_el.text or ""
+            if error_el is not None:
+                status = "error"
+                error_message = error_el.get("message", "")
+                error_text = error_el.text or ""
+            if skipped_el is not None:
+                status = "skipped"
+                skipped_message = skipped_el.get("message", "")
```
cblecker: These are `if` rather than `elif`, so the last match wins. If a testcase has both a `<failure>` and a `<skipped>` element, this reports it as "skipped" — silently hiding a real failure.

I'd either switch to `elif` throughout, or if you want to capture data from all elements, separate the data extraction from the status determination so you can set an explicit priority (error > failed > skipped > passed).

plugins/ci/skills/parse-junit/parse_junit.py
```diff
@@ -0,0 +1,556 @@
...
+# ---------------------------------------------------------------------------
+# Filtering
+# ---------------------------------------------------------------------------
+
+def filter_results(results, name_pattern=None, status_filter=None,
+                   lifecycle_filter=None):
+    """Filter results by name regex, status, and/or lifecycle."""
+    filtered = results
+    if name_pattern:
+        regex = re.compile(name_pattern, re.IGNORECASE)
```
cblecker: If someone passes a bad regex pattern (e.g. `--filter "["`), this will crash with a raw `re.error` traceback. Might be worth wrapping in a try/except with a friendly error message, especially since this runs *after* all files have already been parsed — so you'd lose all the work.

plugins/ci/skills/parse-junit/parse_junit.py
```diff
@@ -0,0 +1,556 @@
...
+
+def parse_file(filepath: str) -> list:
+    """Parse a JUnit XML file, handling gzip transparently."""
+    p = Path(filepath)
+    if not p.exists():
+        print(f"Error: File not found: {filepath}", file=sys.stderr)
+        return []
+
+    if p.suffix == ".gz" or filepath.endswith(".xml.gz"):
+        with gzip.open(filepath, "rt", encoding="utf-8", errors="replace") as f:
```
cblecker, line 229: `gzip.open()` decompresses lazily, so if the .gz file is corrupt the exception (`BadGzipFile`, `zlib.error`, `EOFError`) gets raised inside `ET.parse()` — but the `except ET.ParseError` handler won't catch it. A single bad .gz file in a batch would crash the whole run. Could wrap this in a broader try/except, or catch `OSError` alongside `ParseError` in `parse_junit_xml`.

plugins/ci/skills/parse-junit/parse_junit.py
```diff
@@ -0,0 +1,556 @@
...
+        parser.error("Provide file paths or use --stdin")
+
+    all_results = []
+
+    # Read from stdin
+    if args.stdin:
+        data = sys.stdin.buffer.read()
+        # Transparently decompress gzip
+        if data[:2] == b"\x1f\x8b":
+            data = gzip.decompress(data)
```
cblecker: Same theme as the .gz file handling — `gzip.decompress()` here is outside any try/except. Corrupt or truncated gzip data on stdin (not uncommon when piping from a flaky `gcloud storage cat`) would crash with a traceback instead of a clean error.

plugins/ci/skills/parse-junit/parse_junit.py
```diff
@@ -0,0 +1,556 @@
...
+            }
+        output.append(d)
+    return output
+
+
+# ---------------------------------------------------------------------------
+# CLI
+# ---------------------------------------------------------------------------
+
+def main():
```
cblecker, line 455: Nit: `main()` always exits 0, even if every file fails to parse. The SKILL.md documents exit code 0 = success and 2 = argument error, but there's no exit code for "ran but encountered errors." Might not matter much since this is primarily consumed by AI agents, but if anyone uses it in a `set -e` script they'd get silent failures.

plugins/ci/skills/parse-junit/parse_junit.py
```diff
@@ -0,0 +1,556 @@
...
+            # System-out --------------------------------------------------------
+            sysout_el = tc.find("system-out")
+            system_out = (sysout_el.text or "") if sysout_el is not None else ""
+
+            # Parse aggregated YAML from system-out
+            agg_passes = []
+            agg_failures = []
+            agg_skips = []
+            if system_out and any(
+                k in system_out for k in ("passes:", "failures:", "skips:")
```
cblecker: Minor: this is a substring match, so system-out text that happens to contain "failures:" in a log message (e.g. `"checking for failures: none found"`) would trigger the YAML parser on non-YAML content. Probably fine in practice, but a line-start-anchored regex like `re.search(r"^(?:passes|failures|skips):", system_out, re.MULTILINE)` would be more precise if you wanted to tighten it up.

plugins/ci/skills/parse-junit/SKILL.md
```diff
@@ -0,0 +1,343 @@
...
+
+```bash
+python3 plugins/ci/skills/parse-junit/parse_junit.py junit*.xml \
+  --filter "BackendDisruption" --status failed --format json
+```
+
+### Example 5: Pipe from GCS
+
+```bash
+gcloud storage cat gs://test-platform-results/logs/{job_name}/{build_id}/artifacts/{target}/openshift-e2e-test/artifacts/junit/junit_e2e_*.xml.gz | \
```
cblecker: Heads up — if this glob matches multiple files, `gcloud storage cat` will concatenate them. The decompressed result would be multiple XML documents back-to-back, which isn't valid XML. `ET.parse()` would raise a `ParseError` that gets caught, and you'd silently get empty results. This example should probably either note that it expects a single file match, or use `--stdin` only with a specific file path.

plugins/ci/skills/parse-junit/SKILL.md
```diff
@@ -0,0 +1,343 @@
...
+- `lifecycle`: `blocking` or `informing` — auto-detected from filename/suite, or set via `--lifecycle`
+- `source_file`: Path of the JUnit XML file this test came from
+- `failure_message`: Short failure description (from `<failure message="...">`)
+- `failure_text`: Full failure output (from `<failure>` element text content)
+- `error_message` / `error_text`: Same as failure fields but for `<error>` elements
+- `skipped_message`: Reason for skipping (from `<skipped message="...">`)
+- `system_out`: Raw `<system-out>` content
+- `aggregated`: Only present for aggregated JUnit — contains per-run pass/fail/skip data with Prow URLs
+
+Fields with empty values are omitted from JSON output to reduce noise.
```
cblecker: This isn't quite accurate — `_to_json()` always includes `name`, `status`, `suite_name`, `classname`, `time_seconds`, `lifecycle`, and `source_file` even when empty/zero. Only the optional fields (`failure_message`, `failure_text`, `error_message`, `error_text`, `skipped_message`, `system_out`, `aggregated`) are conditionally omitted. Worth clarifying so consumers know what to expect.

plugins/ci/skills/prow-job-analyze-test-failure/SKILL.md
```diff
@@ -131,13 +131,22 @@ The aggregated junit XML contains links to every underlying job run. Download it
...
-Use the `humanurl` links to investigate individual job run failures with the normal
+```bash
+python3 plugins/ci/skills/parse-junit/parse_junit.py \
+  .work/prow-job-analyze-test-failure/{build_id}/logs/junit-aggregated.xml \
+  --format json
+```
+
+Each test result in the JSON output includes an `aggregated` field with `passes`, `failures`,
+and `skips` lists. Each entry has:
+- `jobRunID`: The build ID of the underlying job run
```
cblecker: The YAML parser preserves field names as-is from the source without normalizing casing. I notice the formatter code in `parse_junit.py` (around line 392) defensively checks both `humanURL` and `humanUrl`, and both `jobRunID` and `jobrunid` — which suggests the upstream aggregator's casing isn't guaranteed. Might be worth either normalizing in the parser, or adding a note here that casing depends on the upstream aggregator.

cblecker (commented):
Nice addition — having a reusable JUnit parser instead of the agent reinventing XML parsing inline every session is a clear win. The code is clean, the SKILL.md is thorough, and the integration with `prow-job-analyze-test-failure` is well done.

The main thing I'd want fixed before merge is the status precedence issue (the `if`/`if`/`if` chain at line 169-179) — that one could silently hide test failures, which is the opposite of what this tool should do. The error handling around gzip files could also use some hardening so a single corrupt file doesn't crash the whole run.

Smaller comments inline on the docs and a couple of other edge cases.

stbenjam:
I am going to take a different approach

PR #424, opened 2026-04-20
Switching the snowflake plugin to use the mart that everyone at Red Hat has access to.

brenton:
## Summary by CodeRabbit

* **New Features**
  * Added --uncategorized flag to filter reports for items missing an Activity Type.

* **Bug Fixes & Improvements**
  * Refined default "completed work" filtering and switched documented data sources to PII-sanitized views for improved privacy.

* **Documentation**
  * Updated Snowflake setup guidance and command docs, including examples and notes about view vs. table usage.

* **Chores**
  * Plugin version bumped to 0.4.0.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -50,7 +50,7 @@
       "name": "snowflake",
       "source": "./plugins/snowflake",
       "description": "Snowflake data analysis commands for engineering metrics and reports",
-      "version": "0.3.0"
+      "version": "0.4.0"
     },
     {
       "name": "sosreport",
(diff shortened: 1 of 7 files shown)


brenton:
I'm not going to address the first nit.  This entire thing is Red Hat specific.  The second nitpick seems reasonable though so I just pushed an update:

Verify each finding against the current code and only fix it if needed.

In `@plugins/snowflake/commands/activity-type-report.md` at line 3, Update the
synopsis section in the activity-type-report documentation to include additional
example usages that show --uncategorized combined with --todo and --all;
specifically add lines demonstrating "/snowflake:activity-type-report <projects>
[months] --uncategorized --todo" and "/snowflake:activity-type-report <projects>
[months] --uncategorized --all" (also mirror the same examples where similar
synopsis lines appear around the later occurrence referenced by the reviewer) so
readers can immediately see flag compatibility with --uncategorized, --todo, and
--all while keeping the existing argument-hint intact.

PR #499, opened 2026-05-24
Replace curl|bash with verified Claude Code install in Dockerfile

stbenjam:
## Summary
- Replaces `curl -fsSL https://claude.ai/install.sh | bash` in `images/Dockerfile` with Claude Code's documented integrity verification flow
- Pins to a specific version (`CLAUDE_CODE_VERSION` build arg, default `2.1.150`)
- Imports Anthropic's GPG key and verifies its fingerprint
- Downloads the signed `manifest.json` and verifies the GPG signature
- Downloads the binary and validates its SHA-256 against the manifest
- To bump the version, update the `CLAUDE_CODE_VERSION` ARG default

See: https://code.claude.com/docs/en/setup#binary-integrity-and-code-signing

## Test plan
- [ ] Build the container image and verify `claude --version` returns the pinned version
- [ ] Tamper with the expected fingerprint to verify the build fails

## Summary by CodeRabbit

* **Chores**
  * Improved CLI installation in the Docker image to use a signed package repository with GPG verification for stronger supply-chain security.
  * Removed a legacy PATH environment adjustment for the CLI.
  * File and directory permissions are now applied after switching to the non-root runtime user to reduce privileged operations during build.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -27,6 +27,17 @@ RUN dnf install -y \
     pcp-export-pcp2json \
     && dnf clean all
 
+# Install Claude Code CLI via official RPM repository
+RUN tee /etc/yum.repos.d/claude-code.repo <<'EOF'
+[claude-code]
+name=Claude Code
+baseurl=https://downloads.claude.ai/claude-code/rpm/stable
+enabled=1
+gpgcheck=1
+gpgkey=https://downloads.claude.ai/keys/claude-code.asc
+EOF
+RUN dnf install -y claude-code && dnf clean all
+
 # Create claude user with root group for OpenShift compatibility
 RUN useradd -m -u 1000 -g 0 -s /bin/bash claude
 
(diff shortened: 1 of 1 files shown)


images/Dockerfile
```diff
@@ -48,8 +48,39 @@ WORKDIR /workspace
...
+    curl -fsSL -o claude "$REPO/$CLAUDE_CODE_VERSION/$PLATFORM/claude"; \
+    ACTUAL_SHA=$(sha256sum claude | cut -d' ' -f1); \
+    [ "$ACTUAL_SHA" = "$EXPECTED_SHA" ]; \
+    \
+    chmod +x claude; \
+    mkdir -p /home/claude/.local/bin; \
+    mv claude /home/claude/.local/bin/claude; \
+    rm -f manifest.json manifest.json.sig
 
 # In CI, OpenShift uses a random UID, but in the root group
```
cblecker: Since the base image is RHEL 9 and this Dockerfile already uses `dnf` for everything else (Python, GitHub CLI, PCP), consider using the official Claude Code RPM repository instead of manual GPG + SHA-256 verification. It's documented at https://code.claude.com/docs/en/setup#dnf:

```dockerfile
RUN tee /etc/yum.repos.d/claude-code.repo <<'EOF'
[claude-code]
name=Claude Code
baseurl=https://downloads.claude.ai/claude-code/rpm/stable
enabled=1
gpgcheck=1
gpgkey=https://downloads.claude.ai/keys/claude-code.asc
EOF
RUN dnf install -y claude-code && dnf clean all
```

This would:
- **Reduce this from ~33 lines to ~10** while achieving the same security goal (`gpgcheck=1` handles signature verification automatically)
- **Be consistent** with how GitHub CLI and other packages are already installed in this Dockerfile
- **Eliminate the pinned version** — no `CLAUDE_CODE_VERSION` ARG to manually bump; `dnf upgrade claude-code` handles it
- **Install to a standard location** (`/usr/bin/claude`) instead of `~/.local/bin`, simplifying the `PATH` setup

The tradeoff is losing the ability to pin an exact version via the build arg, but the `stable` channel already provides a ~1 week delay from bleeding edge, which seems sufficient for CI image builds.

cblecker (commented):
Nice improvement over `curl | bash`. Since this is a RHEL 9 image that already uses `dnf` throughout, the official Claude Code RPM repo might be an even simpler path — see inline comment.

images/Dockerfile
stbenjam: Heh, I totally missed they were publishing RPM's now.  Anthropic is really growing up.

images/Dockerfile
```diff
@@ -27,6 +27,17 @@ RUN dnf install -y \
...
+# Install Claude Code CLI via official RPM repository
+RUN printf '%s\n' \
+    '[claude-code]' \
+    'name=Claude Code' \
+    'baseurl=https://downloads.claude.ai/claude-code/rpm/stable' \
+    'enabled=1' \
+    'gpgcheck=1' \
+    'gpgkey=https://downloads.claude.ai/keys/claude-code.asc' \
+    > /etc/yum.repos.d/claude-code.repo && \
+    dnf install -y claude-code && dnf clean all
```
cblecker, line 39: I was originally thinking we could use `dnf config-manager --add-repo` here to match the gh-cli pattern above, but it turns out Anthropic doesn't publish a `.repo` file URL — their [docs](https://code.claude.com/docs/en/setup#install-with-linux-package-managers) have you create the file manually. So the inconsistency with gh-cli has a legit reason.

That said, since this is static config that won't change between builds, what do you think about moving it to `images/claude-code.repo` and using `COPY` — same pattern as `claude-settings.json` and `known_marketplaces.json`? Would keep the Dockerfile a bit cleaner and make the repo config easier to read and diff separately.

Something like:
```dockerfile
COPY images/claude-code.repo /etc/yum.repos.d/claude-code.repo
RUN dnf install -y claude-code && dnf clean all
```

cblecker (commented):
The switch from curl|bash to the RPM repo is a nice improvement — good stuff.

A couple of things:

1. **PR title and description are stale** — the title still says "verified Claude Code install" and the body describes the original approach (importing GPG keys, verifying manifest signatures, validating SHA-256 checksums, pinning to `CLAUDE_CODE_VERSION`). None of that applies to the current implementation. The test plan also references fingerprint tampering and pinned versions. Could you update these to match what the code actually does now?

2. See the inline comment about the repo file — suggesting moving it to a static file in `images/` for consistency with the other config files.

PR #516, opened 2026-06-02
Fix node-cve Jira report footer

saschagrunert:
- Link to the plugin README instead of the repo root
- Prevent the date from being appended to the footer (the Jira comment timestamp already covers it)

## Summary by CodeRabbit

* **Documentation**
  * Updated the Jira comment template footer with revised instructions for proper usage.
  * Modified the attribution link in the footer to reference a more specific location.
  * Added clarification that date information should not be appended to the footer, as Jira automatically timestamps entries.


saschagrunert:
@bitoku @sairameshv @qiwang19 PTAL

PR #518, opened 2026-06-02
Install Claude Code via DNF instead of curl|bash

cblecker:
## Summary

- Replace the `curl | bash` Claude Code installer with Anthropic's official signed DNF repository
- Move all external `.repo` files (Claude Code and GitHub CLI) into `images/yum.repos.d/` and COPY them in, rather than fetching URLs or inlining heredocs in the Dockerfile
- Set `ART_DNF_WRAPPER_POLICY=append` so the ART yum/dnf wrapper picks up the copied repos alongside its own, enabling local builds on both x86_64 and ARM64

Supersedes #499 

## Test plan

- [x] Build the image locally with `podman build -f images/Dockerfile .`
- [x] Run `podman run --rm <image> --version` and confirm `claude` is present and returns a version
- [x] Verify CI image build succeeds

## Summary by CodeRabbit

* **Chores**
  * Updated container image repositories to include Claude Code and GitHub CLI packages with GPG signature verification enabled
  * Streamlined package installation process by removing inline installation scripts and consolidating repository configuration


images/yum.repos.d/claude-code.repo
```diff
@@ -0,0 +1,6 @@
+[claude-code]
+name=Claude Code
+baseurl=https://downloads.claude.ai/claude-code/rpm/stable
+enabled=1
+gpgcheck=1
+gpgkey=https://downloads.claude.ai/keys/claude-code.asc
```
stbenjam: This largely collapses the GPG trust boundary, we might want to include the GPG keys here and not fetch them from the web.

With the key here, its not really that safer than `curl | bash` since you're trusting `downloads.claude.ai` to be safe.  Granted, the risk is small but it is honestly small for `curl | bash`

(Same for GH below)

PR #534, opened 2026-06-09
Update skillsaw to 0.13.1

not-stbenjam:
## Summary
- Update skillsaw from 0.11.4 to 0.12.1 across config, Makefile, and all GitHub Actions (pinned to commit SHA)
- Enable new supply chain protection rules: `hooks-dangerous`, `hooks-prohibited`, and `settings-dangerous`
- Extract native-notifications inline shell commands into `hooks/scripts/notify.sh` to clean up the `hooks-prohibited` allowlist

## Test plan
- [ ] `make lint` passes with no errors
- [ ] Native notifications still fire on macOS (`osascript`) and Linux (`notify-send`)

## Summary by CodeRabbit

* **Chores**
  * Bumped tooling to skillsaw v0.13.1 across CI, workflows, and Makefile; updated workflow steps to pinned v0.13.1.
  * Strengthened supply-chain rules in configuration and added an explicit allowlist for approved hook commands.

* **Refactor**
  * Unified native notifications into a single cross-platform handler script; plugin bumped to v0.0.3 and author metadata updated.

* **Documentation**
  * Updated plugin docs, README examples, and marketplace entry to reflect the new notification implementation.


stbenjam:
/ok-to-test
/approve

(Automated: Extending self-approval rights and ok-to-test to my bot account.)

plugins/native-notifications/hooks/hooks.json
```diff
@@ -16,7 +16,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "if command -v osascript >/dev/null 2>&1; then osascript -e 'display notification \"Claude finished your task\" with title \"✅ Claude Code\"'; elif command -v notify-send >/dev/null 2>&1 && [ -n \"${DISPLAY:-}${WAYLAND_DISPLAY:-}\" ]; then notify-send '✅ Claude Code' 'Claude finished your task'; else printf '\\a'; fi"
+            "command": "bash ${CLAUDE_PLUGIN_ROOT}/hooks/scripts/notify.sh '✅ Claude Code' 'Claude finished your task'"
```
smith-xyz, line 19: @stbenjam notify.sh resolves osascript via command -v rather than absolute path, does Claude Code sanitize PATH/BASH_ENV before spawning shell-form hooks, or could another plugin shadow the binary?

```bash
mkdir -p /tmp/shadow
printf '#!/bin/bash\nenv | grep -iE "secret" >&2\n/usr/bin/osascript "$@"' > /tmp/shadow/osascript
chmod +x /tmp/shadow/osascript
PATH="/tmp/shadow:$PATH" bash notify.sh '🔔 Claude Code' 'Claude needs your input'
```

something like that if you will

stbenjam: I think this is a general problem, if you can change $PATH you can already do bad things including shadowing claude itself.  You don't need to inject hooks to do bad things.

I only touched this because I didn't care for how the whole complex bash thing was in the hook.

plugins/native-notifications/hooks/scripts/notify.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
```
cblecker: This introduces a subtle behavioral change from the old inline code. The old `if/elif/else/fi` one-liner always exited 0 regardless of whether `osascript` or `notify-send` actually succeeded — but with `set -e`, if `osascript` fails (macOS Focus/DND mode, scripting bridge timeout, sandboxed environment), the script exits non-zero and Claude Code shows a "hook error" label in the transcript. For a best-effort notification feature, that's going to be noisy and confusing for users.

Also worth noting: `set -e` breaks the fallback behavior on execution failure. If `osascript` is present but broken, the script dies before reaching `notify-send` or the terminal bell.

I'd suggest dropping to just `set -u` here and adding `|| true` after the `osascript` and `notify-send` calls — keeps the undefined-variable protection but ensures the script always exits cleanly, matching the old behavior.

plugins/native-notifications/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "native-notifications",
   "description": "Cross-platform desktop notifications (macOS, Linux desktop, Linux headless)",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
cblecker, line 4: nit: while you're in here, mind updating `author.name` on line 6 from `"openshift"` to `"github.com/openshift-eng"`? All the other plugins use that value and the contributing guide requires it.

cblecker (commented):
A couple more minor things:

- The README still shows the old inline `if/elif/else/fi` pattern in the "Hook Configuration" section — should be updated to reflect the new `notify.sh` invocation.
- The PR description mentions "0.12.0" but the code consistently uses 0.12.1 throughout — just a stale description.

not-stbenjam:
Pushed fixes for all three items:

- `notify.sh`: switched to `set -u` and added `|| true` after `osascript`/`notify-send` so notifications stay best-effort
- `plugin.json`: updated `author.name` to `github.com/openshift-eng`
- README: replaced the old inline pattern with the `notify.sh` invocation, and fixed the stale `macos-notifications` path reference
- PR description: corrected 0.12.0 → 0.12.1

plugins/native-notifications/README.md
```diff
@@ -27,7 +27,7 @@ To customize the contents displayed by the notification, edit the script call:
         "hooks": [
           {
             "type": "command",
-            "command": "if command -v osascript &>/dev/null; then osascript -e 'display notification \"Claude needs your input\" with title \"🔔 Claude Code\"'; elif command -v notify-send &>/dev/null && [ -n \"${DISPLAY:-}${WAYLAND_DISPLAY:-}\" ]; then notify-send '🔔 Claude Code' 'Claude needs your input'; else printf '\\a'; fi"
+            "command": "${PLUGIN_DIR}/scripts/notify.sh '🔔 Claude Code' 'Claude needs your input'"
```
cblecker: The example here doesn't match the actual hook command in hooks.json — it uses `${PLUGIN_DIR}/scripts/notify.sh` but the real command is `bash ${CLAUDE_PLUGIN_ROOT}/hooks/scripts/notify.sh`. Three things are off: the variable name (`PLUGIN_DIR` vs `CLAUDE_PLUGIN_ROOT`), the missing `bash` prefix, and the path (`scripts/` vs `hooks/scripts/`). Someone copy-pasting this to customize their notifications would get a broken hook.

plugins/native-notifications/hooks/scripts/notify.sh
```diff
@@ -0,0 +1,17 @@
...
+if command -v osascript >/dev/null 2>&1; then
+  escaped_title="${title//\\/\\\\}"
+  escaped_title="${escaped_title//\"/\\\"}"
+  escaped_message="${message//\\/\\\\}"
+  escaped_message="${escaped_message//\"/\\\"}"
+  osascript -e "display notification \"$escaped_message\" with title \"$escaped_title\"" || true
+elif command -v notify-send >/dev/null 2>&1 && [ -n "${DISPLAY:-}${WAYLAND_DISPLAY:-}" ]; then
+  notify-send "$title" "$message" || true
+else
+  printf '\a'
```
cblecker: nit: `printf '\a'` is missing `|| true` — the other two branches have it for consistency and to avoid accidentally propagating a non-zero exit on a Stop hook. Extremely unlikely to matter, but trivial to fix.

cblecker (commented):
The skillsaw bump and supply chain rules look good. Two small things on the notification refactor — the README example doesn't match the actual hook command, and the printf fallback is missing `|| true` for consistency with the other branches.

stbenjam:
Addressed @cblecker's latest feedback and bumped skillsaw to 0.13.1:

- **README**: fixed hook example to use `bash ${CLAUDE_PLUGIN_ROOT}/hooks/scripts/notify.sh` matching the actual `hooks.json`
- **notify.sh**: added `|| true` to `printf '\a'` fallback for consistency with `osascript`/`notify-send` branches
- **skillsaw**: bumped from 0.12.1 → 0.13.1 across all workflows, Makefile, and `.skillsaw.yaml`

PR #541, opened 2026-06-12
ci:payload-analysis: drop command wrapper, let skill be invoked directly

petr-muller:
The command file duplicated enough of the skill's content that the agent treated it as the full spec and never opened SKILL.md — observed in practice when RHCOS badges, variant isolation, and verbatim subagent prompts were all skipped.

Skills are directly invocable, so the command wrapper was unnecessary. Removing it eliminates the failure mode entirely. Examples and argument-hint migrated into SKILL.md.

## Summary by CodeRabbit

* **Documentation**
  * Updated the payload-analysis skill documentation with a clearer directive for the required implementation steps.
  * Added an argument hint for the CLI invocation format and a new Examples section with practical usage scenarios.
* **Chores**
  * Bumped the CI plugin version to **0.0.49** consistently across marketplace metadata and the public docs/plugin manifests.


plugins/ci/commands/payload-analysis.md
stbenjam, line 1: We should drop this entire `/` command entirely, skils are `/` invocable.

petr-muller: good point, addressed

plugins/ci/skills/payload-analysis/SKILL.md
```diff
@@ -21,6 +22,23 @@ Use this skill when you need to:
...
+
+2. **Analyze using an existing snapshot directory**:
+   ```
+   /ci:payload-analysis 4.22.0-0.nightly-2026-02-25-152806 --snapshot-dir payload/4.22/nightly
+   ```
+
+3. **Analyze an arm64 payload** (architecture inferred from tag):
+   ```
+   /ci:payload-analysis 4.22.0-0.nightly-arm64-2026-02-25-152806
+   ```
```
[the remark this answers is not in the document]
petr-muller: these are not bash and they are single-line commands so even syntax highlight does not have great valuee

PR #554, opened 2026-06-17
openshift-developer: add hooks, skills, and workflow documentation

enxebre:
## Summary

- Add SessionStart hook that installs pre-commit and pre-push hooks via `pre-commit` when a repo has `.pre-commit-config.yaml`
- Add `address-review-pr` skill (moved from `utils:address-reviews`) — fetches and addresses PR review comments by priority
- Add `address-review-precommit` skill — fixes code review findings before committing
- Add `git-commit-format` skill — conventional commit formatting rules (ported from openshift/hypershift)
- Document pre-PR and post-PR workflows in README
- Add prerequisites section (pre-commit, gitlint, gopls, gh)
- Bump version to 1.1.0

## Test plan

- [ ] Install plugin: `claude plugin install openshift-developer@ai-helpers`
- [ ] Verify `/openshift-developer:address-review-pr` is available
- [ ] Verify `/openshift-developer:address-review-precommit` is available
- [ ] Verify `/openshift-developer:git-commit-format` is available
- [ ] Verify SessionStart hook runs in a repo with `.pre-commit-config.yaml`
- [ ] `make lint` passes

## Summary by CodeRabbit

* **New Features**
  * Added OpenShift Developer skills to help address PR review feedback and run review-precommit fixes.
  * Introduced a `git-commit-format` skill for Conventional Commit-style messages.
  * Added an automatic pre-commit hook setup step when a pre-commit config is present.
* **Chores**
  * Bumped plugin versions (git, utils, OpenShift Developer) and updated OpenShift Developer’s Jira dependency range.
* **Removals**
  * Removed the `utils:address-reviews` command documentation.


plugins/openshift-developer/scripts/ensure-precommit.sh
stbenjam, line 1: I am concerned about this --https://pre-commit.com/ was a nice idea when the world was safer, but this means any repo cloned off the internet will have its pre-commit hooks installed, and then executed on a commit.  That's a significant trust boundary to cross and not one we should turn on by default. 

I think its something developers should opt-into if they want to use it, and run `pre-commit install` themselves on repos they trust.

stbenjam: That said, we should probably turn on https://github.com/leaktk -- the rh precommit hook from prodsec I think is available on the internet now, and everyone should have this

stbenjam (changes requested):

plugins/utils/commands/address-reviews.md
bryan-cox, line 1: How does this affect the periodic and the GHA we have now that runs this plugin?

plugins/openshift-developer/scripts/ensure-precommit.sh
enxebre: I share the concerns. Though installing precommit is still a deliberate prerequisite https://github.com/openshift-eng/ai-helpers/pull/554/changes#diff-9ef7a0f3f7427ef5b05b58ed94e126aaf2ba228c968ea2621c0e6cf27e2a79adR4-R9.

The intent is to make this bundle opinionated and as deterministic as possible so it produces measurable consistent quality output.
Running our curated and prescribed precommit/push hooks would not be a choice for consumers but a prescription enforced by the executable workflow. That I'll want to remain true while adjusting from a security standpoint.

The main driving use case atm is ocpbug autofix and local dev for hcp where this approach is convenient. We could possibly make this work via run via claude hooks instead.

In future I would expect we distribute this bundle in a preconfigured sandboxed environment where we could also add prescriptive hooks/scripts

plugins/utils/commands/address-reviews.md
enxebre: The intent is to make this bundle the main way we distribute the executable dev workflows everywhere. They'll need to be updated

plugins/openshift-developer/scripts/ensure-precommit.sh
stbenjam: Installing pre-commit as a tool is materially different than running `pre-commit install` in every repository you ever clone and use in Claude, which is what the ensure-precommit hook does if I'm reading this right

stbenjam: @coderabbitai What do you think of this discussion?

enxebre: added Allowlist

plugins/openshift-developer/skills/address-review-precommit/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+- If `make verify` generates new files, commit those too and run `make verify` again to confirm it passes
+- Maximum 3 retry attempts if verification fails — fix the issues and re-run
+- If verification still fails after 3 attempts, stop and report to the user
+
+### Step 4: Commit and push
+
+1. Amend existing commits or create new commits as appropriate
+2. Push the branch to origin:
+   ```bash
+   git push --force-with-lease
```
stbenjam: +1 to coderabbit here, I try to keep my claude from ever force pushing, I can use tide's autosquash to squash things

plugins/openshift-developer/scripts/ensure-precommit.sh
stbenjam: Thanks, this resolves my comment here

plugins/git/skills/git-commit-format/SKILL.md
stbenjam, line 1: I am not 100% sure these skills belong here, should this remain as a meta-module that pulls in others?

plugins/openshift-developer/skills/address-review-precommit/SKILL.md
enxebre: the original intent was for this skills to ammend commits where it make sense, hence the need to push force.
I can also see value on a workflow that always commit/push atop and let the commit/squash be a separate step. Dropped the --force  flag. We can adjust and introduce other opinions as we iterate.

plugins/git/skills/git-commit-format/SKILL.md
enxebre: Me neither, I put it here initially instead of in the git plugin to avoid bringing many of those specs which are unneeded here as a dependency. This skill has a concrete use case and as a principle seems easier to iterate with less deps.
I moved it there and added the dependency with the git plugin. I would expect eventually we enforce distribution of a .gitlint file that runs on precommit hook

PR #556, opened 2026-06-18
`ci:payload-analysis`: detect CI step-registry changes as revert candidates

smg247:
Add Step 3.6 to the payload-analysis skill to identify CI infrastructure changes (step-registry modifications in openshift/release) that correlate with failure streaks. Uses the unauthenticated GitHub Commits API to find step-registry changes around the originating payload date, then triages results against failure context.

Add eval [case-014](https://redhat.atlassian.net/browse/case-014) testing this capability against the PR #78140 mirror URL migration that broke metal-ipi gather-must-gather.

## Summary by CodeRabbit

* **Chores**
  * Bumped the CI plugin version to **0.0.52** across marketplace metadata and the site.
* **New Features**
  * Enhanced payload analysis with a new **Step 3.6** to correlate failed CI jobs with recent step-registry changes.
* **Tests**
  * Updated the payload-analysis evaluation case (new payload tag and expected outcome) to reflect the latest scenario.
* **Documentation**
  * Expanded the payload-analysis workflow guide, including **Step 3.6** and required **GitHub CLI (`gh`)** steps.


plugins/ci/skills/payload-analysis/SKILL.md
```diff
@@ -125,6 +125,53 @@ From `summary.json` → `test_failures.blocking[]`:
 
 For deeper context, read `build_log.json` (at the `build_log` path) for any failed job. It contains `error_warning_lines[]` with `line_number` and `text`, plus `tail_lines[]` (last 20% of the log).
 
+#### 3.6: Check for CI Infrastructure Changes (Streak >= 3 Only)
```
stbenjam: Did you try it without a streak to see if it could get it on the first try (without going off the rails for other cases)

plugins/ci/skills/payload-analysis/SKILL.md
```diff
@@ -125,6 +125,53 @@ From `summary.json` → `test_failures.blocking[]`:
...
+Extract the date from the `originating_payload` tag (format: `<version>-0.<stream>-YYYY-MM-DD-HHMMSS`). Convert the timestamp to an ISO date for the GitHub API (e.g., `2026-06-16` from `5.0.0-0.nightly-2026-06-16-185706`). Compute a time window: `since` = originating date minus 1 day, `until` = originating date plus 1 day (ISO 8601 format with `T00:00:00Z` suffix).
+
+Use the GitHub Commits API (no authentication required) to find step-registry changes around the originating payload date.
+
+**First, get all step-registry commits in the time window:**
+
+```bash
+curl -s "https://api.github.com/repos/openshift/release/commits?path=ci-operator/step-registry&since=<since_date>T00:00:00Z&until=<until_date>T23:59:59Z&per_page=50" \
+    | jq '.[] | {sha: .sha[0:11], date: .commit.committer.date, message: .commit.message | split("\n")[0]}'
+```
```
stbenjam, line 139: `gh` is available in our container image and authenticated while in CI, the curls could get rate limited in CI if not authed

smg247: ah, I was wondering that. I had tried it locally with just case 014 and it did terribly because GH wasn't available. I brought that up to Claude, and it told me that it wouldn't be available in the container either so we pivoted. I can go back to using `gh`, and try it again here.

plugins/ci/skills/payload-analysis/SKILL.md
smg247: I didn't. Is it worth trying a run without the streak requirement with the full analysis again?

plugins/ci/skills/payload-analysis/SKILL.md
stbenjam: Claude is confused :) https://github.com/openshift-eng/ai-helpers/blob/1015dd8ae47d5ff4f0715d6fc79a95e56e1ec9ca/images/Dockerfile#L25

smg247:
/hold the analysis is suggesting that the GH cli isn't available. I am investigating. Here is what Claude believes the issue is:

> [Case-001](https://redhat.atlassian.net/browse/Case-001) expectations: The annotations specifically expect the agent to find openshift/release#76852 as a CI config issue affecting install-analysis-all. PR #76852 modified
  ci-operator/step-registry/ipi/conf/vsphere/check/vcm/ — so our Step 5b path filter would find it. I verified the gh api query works locally and returns the right commit.

>  What happened in CI: The eval log shows [case-001](https://redhat.atlassian.net/browse/case-001) DID invoke gh api for step-registry commits multiple times. But the agent's analysis says "GitHub CLI was not available to check for step-registry
  changes." [Case-007](https://redhat.atlassian.net/browse/Case-007) is even more explicit: "gh is not authenticated, so let's search via web" and then falls back to curl.

>  Root cause: With parallelism=5 in the eval, multiple cases run simultaneously making gh api calls (including subagents). This likely causes GitHub API rate limiting or token exhaustion, which the agent
  interprets as "gh not available." [Case-014](https://redhat.atlassian.net/browse/Case-014) happened to run when the rate limit wasn't exhausted, so it worked.

>  The bottom line: Step 5b relies exclusively on gh with no fallback, and gh fails intermittently under parallel load in CI. We need a curl fallback for the GitHub API calls.

stbenjam:
Hm, that's suspcious, gh cli is there, and I thought Prow was injecting a GITHUB_TOKEN but I don't see it in the job info.  We'll hit rate limits regardless of gh or the api though.  Maybe we need to explicitly add a GH token?

The snapshots do include PR data, but the new step may be encouraging too much github activity to trip thresholds.

smg247:
https://github.com/openshift/release/pull/80756 should resolve the unauthenticated `gh` usage

plugins/ci/skills/payload-analysis/SKILL.md
```diff
@@ -125,6 +125,50 @@ From `summary.json` → `test_failures.blocking[]`:
...
+    --jq '.[] | {sha: .sha[0:11], date: .commit.committer.date, message: (.commit.message | split("\n")[0])}'
+```
+
+If this finds nothing, also check steps that run earlier in the workflow and set up infrastructure the failing step depends on (e.g., if `openshift-e2e-test` fails due to connectivity, check `baremetalds/devscripts/proxy` or `ipi/conf` steps that configure networking).
+
+**Scoring CI infrastructure candidates.** If a commit/PR modified a step that the failing job executes (or a shared dependency of that step), flag it as a **CI infrastructure candidate** — include it in Step 6.1 scoring alongside component PR candidates. When the failure's error messages reference URLs, domains, binaries, or configurations that were changed by the PR, the error message match signal (+40) should fire strongly. The key test: does the PR's diff introduce, modify, or remove something that appears in the error output?
+
+This step catches failures caused by CI tooling changes (mirror URL migrations, proxy configuration updates, script refactors) that are invisible to the snapshot's PR tracking.
+
```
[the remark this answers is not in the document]
smg247: 1. Prerequisites/GITHUB_TOKEN doc: Skip — gh uses GITHUB_TOKEN automatically when set; that's standard gh behavior, not something this skill needs to document. The skill-level prerequisites already list
   gh on line 53. Adding token setup instructions here would duplicate infra concerns that belong in the CI job config (which we just fixed).
  2. Error handling for 403/401: Skip — this is an LLM skill doc, not a shell script. The model will see the error output and react appropriately. Prescribing error handling paths adds bulk without
  improving the model's behavior, and rate limiting is not a realistic concern with an authenticated token hitting 5,000 req/hr.
  3. Pagination clarity: Valid — the current wording is vague. Will fix.
  4. Streak ambiguity: Valid — the PR commit message mentions "remove streak >= 3 requirement" but someone reading the current code can see line 130 says "for each failed job" with no streak gate. The
  code is correct and unambiguous. Skip — no change needed; the code already says what it means.

PR #560, opened 2026-06-19
feat: add metrics plugin with OpenTelemetry and OpenInference support

kenjpais:
## Summary

- Adds `plugins/metrics/` — a configuration-only plugin that instruments Claude Code CLI sessions with OpenTelemetry tracing, metrics, and log events and optionally enriched with [OpenInference](https://arize-ai.github.io/openinference/) semantic conventions
- Uses the standard OTel Collector tool as a background process with no custom Python scripts. It collects Claude Code OTel telemetry and writes it to JSON file. Metrics collection is handled directly by the collector. All signals come from Claude Code's built-in telemetry. 
- Uses an `otelcol-contrib` pipeline: native `claude_code.*` spans → OpenInference span kinds (`CHAIN`, `LLM`, `TOOL`, `AGENT`) and attributes (`llm.token_count.*`, `llm.model_name`, etc.) → rotating JSONL file; MLflow export included
- Sub-agent telemetry supported: `Task` tool invocations map to `AGENT` span kind with `graph.node.id` / `graph.node.parent_id`

## Test plan

- [ ] Start collector: `CLAUDE_METRICS_LOG_DIR=~/.local/share/claude-metrics otelcol-contrib --config plugins/metrics/config/otelcol.yaml`
- [ ] Open a Claude Code session with the plugin active; run a few tool calls; confirm `claude-metrics.jsonl` contains spans with `openinference.span.kind` and `llm.token_count.*`
- [ ] `make lint`

## Summary by CodeRabbit

* **New Features**
  * Added a metrics plugin enabling telemetry collection and observability for Claude Code
  * Automatically captures traces, metrics, and logs with privacy-aware filtering to protect sensitive data
  * Supports optional MLflow integration for trace export and analysis
  * Stores observability data locally for debugging and monitoring purposes


stbenjam:
We have been using https://github.com/opendatahub-io/agentic-ci for this,  not sure we should have a separate implementation

not-stbenjam:
Tested this against agentic-ci's tracing approach by running both side-by-side and pushing traces to a local MLflow instance. A few findings:

**Comparison with agentic-ci:**

| | ai-helpers (this PR) | agentic-ci |
|---|---|---|
| **Collector** | `otelcol-contrib` — declarative YAML transforms | Python stdlib `http.server` — ~100 lines, zero dependencies |
| **OpenInference** | Yes — span kinds (`CHAIN`/`LLM`/`TOOL`/`AGENT`), `llm.token_count.*`, `llm.model_name`, agent graph IDs | No — raw `claude_code.*` spans passed through unchanged |
| **Input/output extraction** | Can extract `input.value` from span attrs + `output.value` from span events via OTTL (see below) | None |
| **MLflow export** | Real-time via `otlphttp` exporter in the collector pipeline | Post-hoc — reads JSONL, converts to protobuf, POSTs to `/v1/traces` |
| **Privacy filtering** | Configurable `filter/privacy` processor that drops spans containing prompt/tool content | None |
| **Metrics** | Passes through native `claude_code.*` metrics as-is | Parses metrics in Python for live token rate display and post-run summary |

**What works well:**
- The OpenInference span kind mapping and token count attributes come through cleanly in MLflow. This is a meaningful improvement over agentic-ci, which passes raw spans through without any semantic enrichment.
- The `otelcol-contrib` transform approach is nice — all the mapping logic is declarative YAML, no custom code.

**Gaps found during testing:**

1. **MLflow sessions page shows no inputs or outputs.** MLflow populates its input/output columns from `input.value` and `output.value` span attributes. The current transform doesn't set these. For inputs, mapping `user_prompt` → `input.value` on interaction spans and `full_command` → `input.value` on tool spans works. For outputs, it's trickier — Claude Code emits tool results as **span events** (named `tool.output`), not span attributes. You need a `spanevent` context block to extract them:

   ```yaml
   # In the span context:
   - set(span.attributes["input.value"], span.attributes["user_prompt"])  where span.attributes["user_prompt"] != nil
   - set(span.attributes["input.value"], span.attributes["full_command"]) where span.attributes["full_command"] != nil

   # New spanevent context block:
   - context: spanevent
     statements:
       - set(span.attributes["output.value"], attributes["output"])  where name == "tool.output" and attributes["output"] != nil
       - set(span.attributes["output.value"], attributes["content"]) where name == "tool.output" and attributes["content"] != nil
   ```

2. **No assistant response text in Claude Code's telemetry.** The LLM's actual text response is not emitted anywhere — not as a span attribute, span event, or log record. Session-level output will always be empty regardless of transforms. This is a Claude Code limitation, not fixable here.

3. **Privacy filter vs. content visibility.** If someone enables `OTEL_LOG_TOOL_CONTENT=1` to get tool I/O in their traces, the `filter/privacy` processor silently strips it. That's the intended defense-in-depth behavior, but worth documenting that you need to remove `filter/privacy` from the pipeline to actually see content.

4. **MLflow exporter endpoint format.** The commented-out MLflow exporter uses `otlp/mlflow` with `endpoint: http://localhost:5000/api/2.0/mlflow/otlp`. In current otelcol-contrib (v0.154.0+), the `otlphttp` type expects just `endpoint: http://localhost:5000` — it appends `/v1/traces` automatically. The path-in-endpoint form causes an "invalid port" error.

.claude-plugin/marketplace.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ai-helpers",
   "owner": {
-    "name": "openshift-eng"
+    "name": "kenjpais"
```
Prashanth684: revert

Prashanth684:
> We have been using https://github.com/opendatahub-io/agentic-ci for this, not sure we should have a separate implementation

I believe this is more for local development - agentic-ci seems more for the harness and for running in a containerized fashion.

stbenjam:
> > We have been using https://github.com/opendatahub-io/agentic-ci for this, not sure we should have a separate implementation
> 
> I believe this is more for local development - agentic-ci seems more for the harness and for running in a containerized fashion.

I am more concerned about the fan out of many duplicative sets of code that workaround the limitations of Claude's OTEL outputs and also produce different views of the data.   I like the approach here, but 95% of the things here probably belong in a more reusable package or could go straight to otel-contrib's repo

stbenjam:
@kenjpais There's discussion above and a thread on Slack you should read before continuing to iterate on this

kenjpais:
@stbenjam, thanks alot for the feedback. 
I just updated the PR to address issues found during testing.
Yes, I’ve read the Slack thread - I do agree that we should unify trace collection, we can use the OpenInference and the OpenTelemetry GenAI conventions to achieve that I guess, but every harness may have different raw telemetry requiring adapters for each.
However, I agree that this could be moved into a more reusable package or otel-contrib's repo but I think that would require finalizing schema. Alternatively, the transformation YAML could be maintained separately and loaded by the OTel Collector.
The main goal of this PR was to help users better understand claude's actions and the tools it uses, especially with the use of the agentic-docs plugin ie. to measure agentic-docs usage which will help us better understand the value the agentic-docs plugin provides and identify any gaps or areas for improvement.

I still think this PR would be helpful for the agentic-docs plugin though for now.

stbenjam:
Ok can you take it out of draft? I think this is fine to land and we can sort out alignment later. I think whatever we do would end up based on this work

/lgtm

PR #565, opened 2026-06-23
fix: bundle external deps into ai-helpers marketplace

bryan-cox:
## Problem

The `openshift-developer` bundle failed to load in GHA because its transitive
dependencies (`gopls-lsp` from `claude-plugins-official`, `prodsec-skills` from
`RedHatProductSecurity/prodsec-skills`) live in external marketplaces that
consumers don't have added. The `allowCrossMarketplaceDependenciesOn` field only
*permits* cross-marketplace deps — it does NOT auto-add those marketplaces for
consumers.

Additionally, the `git` plugin is at `0.0.7` but `openshift-developer` required
`^0.0.6`. Under semver `0.x` rules, `^0.0.6` means `>=0.0.6 <0.0.7`, so
`0.0.7` is excluded and the bundle fails with:

```
Error: Requires "git@ai-helpers" ^0.0.6, installed 0.0.7
```

## Fix

1. **Bundle external deps into ai-helpers marketplace** — add `gopls-lsp` and
   `prodsec-skills` as marketplace entries with `source` pointers to their
   upstream repos. Consumers now only need the `ai-helpers` marketplace.

2. **Remove cross-marketplace references** — `golang` plugin's `gopls-lsp` dep
   and `openshift-developer`'s `prodsec-skills` dep no longer specify
   `"marketplace"` since they resolve within ai-helpers now.

3. **Fix git dep constraint** — update `openshift-developer`'s git dep from
   `^0.0.6` to `^0.0.7`.

4. **Version bumps** — `openshift-developer` `1.1.1` → `1.1.2`, `golang`
   `0.3.0` → `0.3.1`.

## Evidence

GHA failure from hypershift PR #8809 (merged):
```
Error: dependency-unsatisfied for openshift-developer@ai-helpers
```

Local verification after fix:
```
$ claude plugin install openshift-developer@ai-helpers
✓ openshift-developer 1.1.2 installed (with gopls-lsp, prodsec-skills resolved)
$ claude -p "/openshift-developer:address-review-pr 8809"
✓ Skill recognized and executed
```

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -3,11 +3,30 @@
   "owner": {
     "name": "openshift-eng"
   },
-  "allowCrossMarketplaceDependenciesOn": [
-    "claude-plugins-official",
-    "prodsec-skills"
-  ],
   "plugins": [
+    {
+      "name": "gopls-lsp",
+      "source": {
+        "source": "github",
+        "repo": "anthropics/claude-plugins-official",
+        "path": "plugins/gopls-lsp"
+      },
+      "description": "Go language server for code intelligence and refactoring",
+      "version": "1.0.0",
+      "category": "development",
+      "keywords": ["go", "gopls", "lsp"]
+    },
+    {
+      "name": "prodsec-skills",
+      "source": {
+        "source": "github",
+        "repo": "RedHatProductSecurity/prodsec-skills"
+      },
+      "description": "Security guidance skills for AI coding assistants",
+      "version": "1.0.0",
+      "category": "security",
+      "keywords": ["security", "prodsec"]
+    },
     {
       "name": "git",
       "source": "./plugins/git",
@@ -322,7 +341,7 @@
       "name": "golang",
       "source": "./plugins/golang",
       "description": "Go development tools: gopls MCP server, LSP integration, automatic gofmt formatting, golangci-lint linting, and CVE dependency patching",
-      "version": "0.3.0",
+      "version": "0.3.1",
       "category": "development",
       "keywords": [
         "go",
@@ -396,7 +415,7 @@
       "name": "openshift-developer",
       "source": "./plugins/openshift-developer",
       "description": "Bundle of curated plugins, skills, and MCP servers useful to any OpenShift engineer",
-      "version": "1.1.1",
+      "version": "1.1.2",
       "category": "bundle",
       "keywords": [
         "openshift",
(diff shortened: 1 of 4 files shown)


bryan-cox:
## Why this change is needed

The `openshift-developer` bundle currently depends on `gopls-lsp` (from `claude-plugins-official`) and `prodsec-skills` (from `RedHatProductSecurity/prodsec-skills`). These are cross-marketplace dependencies — consumers must manually add those external marketplaces before installing the bundle, otherwise it fails to load entirely.

This broke HyperShift's GHA CI when we switched from installing individual plugins to the `openshift-developer` bundle. The failing run: https://github.com/openshift/hypershift/actions/runs/28025683087/job/82952751320

The init JSON from that run shows the errors:
```json
"plugin_errors": [
  {
    "plugin": "openshift-developer@ai-helpers",
    "type": "dependency-unsatisfied",
    "message": "Dependency \"prodsec-skills@prodsec-skills\" is not installed — run `claude plugin install prodsec-skills@prodsec-skills`, or check that its marketplace is added"
  },
  {
    "plugin": "golang@ai-helpers",
    "type": "dependency-unsatisfied",
    "message": "Dependency \"gopls-lsp@claude-plugins-official\" is not installed — run `claude plugin install gopls-lsp@claude-plugins-official`, or check that its marketplace is added"
  }
]
```

Since `openshift-developer` failed to load, its skills (including `address-review-pr`) were never registered, and the `/address-review-comments` GHA job returned "Unknown command".

Per the [Claude Code docs](https://code.claude.com/docs/en/plugin-dependencies#depend-on-a-plugin-from-another-marketplace), `allowCrossMarketplaceDependenciesOn` only *permits* cross-marketplace deps — it doesn't auto-add those marketplaces for consumers. Every consumer has to know about and add them manually.

By including `gopls-lsp` and `prodsec-skills` as plugins in the ai-helpers marketplace (with sources pointing to their upstream repos), all deps resolve within a single marketplace. Consumers just need `claude plugin marketplace add openshift-eng/ai-helpers` and everything works.

This also fixes the `git` dep constraint: `^0.0.6` excludes `0.0.7` under semver `0.x` rules (`^0.0.6` = `>=0.0.6 <0.0.7`).

stbenjam:
Hm, is it too difficult to enable the marketplaces in the GHA? It is sort of nice it just works with this approach but we lose the versioning from the upstream sources

bryan-cox:
> Hm, is it too difficult to enable the marketplaces in the GHA? It is sort of nice it just works with this approach but we lose the versioning from the upstream sources

I can but everyone has to do that whenever they use the OpenShift engineering bundle. I wasn't sure what UX we wanted to go with here.

stbenjam:
Hm, the UX I guess is better than any potential consequences of inlining external deps

/lgtm

PR #593, opened 2026-07-01
node-rpm: add RPM package management plugin

saschagrunert:
Adds the `node-rpm` plugin for RPM package management of OpenShift Node team components. Part of the registered Node Team AI Assistant plugin family.

- `/node-rpm:bump` command walks through spec file updates, source downloads, changelog bumps, and Brew builds
- `--vagrant` flag auto-provisions a Fedora libvirt VM with all required tools, reuses it on subsequent runs
- Vendored Vagrantfile (no external gist dependency)
- VPN connectivity check before starting any work
- Handles multiple RHEL branches per OCP version (e.g. RHEL 9 + RHEL 10 for OCP 5.0)
- Proper Release reset on version bumps (sed to 0, then rpmdev-bumpspec increments to 1)
- Annotated tag support via `git ls-remote` with `^{}` dereference
- Reference doc covering the end-to-end RPM workflow, environment setup, scratch build patterns, and VM lifecycle
- Currently supports cri-tools, extensible via the packages table in `rpm-workflow.md`
- Registered in marketplace and node-team plugin family table

## Checklist:

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -340,6 +340,20 @@
       ]
     },
     {
+      "name": "node-rpm",
+      "source": "./plugins/node-rpm",
+      "description": "RPM package management for OpenShift Node team components",
+      "version": "0.0.1",
+      "category": "openshift",
+      "keywords": [
+        "openshift",
+        "node",
+        "rpm",
+        "bump",
+        "cri-tools"
+      ]
+    },
+    {
       "name": "bigquery",
       "source": "./plugins/bigquery",
       "description": "BigQuery analysis utilities",
--- a/docs/index.html
+++ b/docs/index.html
@@ -1957,6 +1957,38 @@ footer a { color: var(--primary); }
       }
     },
     {
+      "name": "node-rpm",
+      "description": "RPM package management for OpenShift Node team components",
+      "version": "0.0.1",
+      "has_readme": true,
+      "commands": [
+        {
+          "name": "bump",
+          "full_name": "node-rpm:bump",
+          "description": "Bump a downstream RPM package to a new upstream version",
+          "description_html": "Bump a downstream RPM package to a new upstream version",
+          "synopsis": "/node-rpm:bump \u003cpackage\u003e \u003cnew-version\u003e [--ocp-version \u003cversion\u003e] [--scratch] [--vagrant]",
+          "body_html": "\u003cp\u003eGuides the user through bumping a downstream RPM package to a new upstream version in Red Hat&#x27;s dist-git. Handles spec file updates, source downloads, changelog bumps, and Brew builds.\u003c/p\u003e\u003cp\u003eThe workflow follows the established pattern for Node team RPM packages (cri-tools as the primary template). Each phase presents its changes to the user for review before proceeding.\u003c/p\u003e\u003cp\u003eWhen \u003ccode\u003e--vagrant\u003c/code\u003e is passed, the entire workflow runs inside a Fedora Vagrant VM that is automatically provisioned with all required tools. This is the recommended approach when rhpkg and related tools are not installed locally.\u003c/p\u003e"
+        }
+      ],
+      "skills": [],
+      "agents": [],
+      "hooks": [],
+      "mcp_servers": [],
+      "rules": [],
+      "category": "openshift",
+      "keywords": [
+        "openshift",
+        "node",
+        "rpm",
+        "bump",
+        "cri-tools"
+      ],
+      "author": {
+        "name": "github.com/openshift-eng"
+      }
+    },
+    {
       "name": "node-team",
       "description": "OpenShift Node team assistant for development, deployment, debugging, and workflow tasks across kubelet, MCO, CRI-O, crun, conmonrs, Kueue operator, Jira, Red Hat KB/support cases, Prometheus, and platform docs.",
       "version": "0.17.1",
--- a/plugins/node-rpm/commands/bump.md
+++ b/plugins/node-rpm/commands/bump.md
@@ -1,0 +1,275 @@
+---
+description: Bump a downstream RPM package to a new upstream version
+argument-hint: "<package> <new-version> [--ocp-version <version>] [--scratch] [--vagrant]"
+---
+
+## Name
+node-rpm:bump
+
+## Synopsis
+```text
+/node-rpm:bump <package> <new-version> [--ocp-version <version>] [--scratch] [--vagrant]
+```
+
+## Description
+
+Guides the user through bumping a downstream RPM package to a new upstream version in Red Hat's dist-git. Handles spec file updates, source downloads, changelog bumps, and Brew builds.
+
+The workflow follows the established pattern for Node team RPM packages (cri-tools as the primary template). Each phase presents its changes to the user for review before proceeding.
+
+When `--vagrant` is passed, the entire workflow runs inside a Fedora Vagrant VM that is automatically provisioned with all required tools. This is the recommended approach when rhpkg and related tools are not installed locally.
+
+## Implementation
+
+### Phase 0: Setup and Argument Parsing
+
+1. **Parse required arguments:**
+   - `<package>` (required): The package name, e.g. `cri-tools`. Must exist in the Packages table in [`../references/rpm-workflow.md`](../references/rpm-workflow.md). If not found, list supported packages and stop.
+   - `<new-version>` (required): The target upstream version, e.g. `1.36.0`
+
+2. **Parse optional arguments:**
+   - `--ocp-version <version>`: Target OCP version. If omitted, prompt the user to specify one.
+   - `--scratch`: Run a scratch build instead of a full build.
+   - `--vagrant`: Run the workflow inside a Vagrant VM.
+
+3. **Check VPN connectivity:**
+   ```bash
+   curl -s --connect-timeout 5 http://download.devel.redhat.com > /dev/null 2>&1 || echo "UNREACHABLE"
+   ```
+   If unreachable, warn the user to connect to the Red Hat VPN and stop.
+
+4. **Validate prerequisites:**
+
+   **Without `--vagrant` (local execution):**
+   ```bash
+   which rhpkg 2>/dev/null || echo "MISSING: rhpkg"
+   which spectool 2>/dev/null || echo "MISSING: spectool"
+   which rpmdev-bumpspec 2>/dev/null || echo "MISSING: rpmdev-bumpspec"
+   klist -s 2>/dev/null || echo "MISSING: valid Kerberos ticket (run kinit)"
+   ```
+   If any tool is missing, print installation instructions and stop.
+
+   **With `--vagrant`:**
+   - Verify `vagrant` is installed locally.
+   - Verify libvirt is available (`virsh --version`).
+   - Create the work directory: `mkdir -p .work/node-rpm`
+   - If `.work/node-rpm/Vagrantfile` does not exist, copy it from the vendored reference:
+     ```bash
+     cp "${CLAUDE_PLUGIN_ROOT}/references/Vagrantfile" .work/node-rpm/Vagrantfile
+     ```
+   - Check if the VM is already running:
+     ```bash
+     cd .work/node-rpm && vagrant status --machine-readable | grep ",state," | grep -q "running"
+     ```
+     If not running, provision and start it:
+     ```bash
+     cd .work/node-rpm && vagrant up
+     ```
+   - Verify VPN connectivity from inside the VM (NAT may not inherit host routes with split-tunnel VPN):
+     ```bash
+     cd .work/node-rpm && vagrant ssh -c "curl -s --connect-timeout 5 http://download.devel.redhat.com > /dev/null 2>&1" || echo "VM cannot reach Red Hat network"
+     ```
+     If unreachable, warn about potential split-tunnel or DNS issues and stop.
+   - Validate prerequisites inside the VM:
+     ```bash
+     cd .work/node-rpm && vagrant ssh -c "which rhpkg && which spectool && which rpmdev-bumpspec"
+     ```
+   - Check Kerberos ticket inside the VM:
+     ```bash
+     cd .work/node-rpm && vagrant ssh -c "klist -s 2>/dev/null" || echo "MISSING: valid Kerberos ticket"
+     ```
+     If no valid ticket, tell the user to authenticate interactively:
+     ```bash
+     cd .work/node-rpm && vagrant ssh
+     # inside the VM:
+     kinit <user>@IPA.REDHAT.COM
+     exit
+     ```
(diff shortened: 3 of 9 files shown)


.claude-plugin/marketplace.json
```diff
@@ -339,6 +339,20 @@
...
+      "version": "0.0.1",
+      "category": "openshift",
+      "keywords": [
+        "openshift",
+        "node",
+        "rpm",
+        "rebase",
+        "cri-tools"
+      ]
+    },
```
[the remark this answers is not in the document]
saschagrunert: This is stale now?

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,261 @@
...
+
+2. **Push** (confirm with the user first):
+   ```bash
+   git push
+   ```
+
+3. **Start the build:**
+   - Full build: `rhpkg build`
+   - Scratch build (if `--scratch`): `rhpkg build --scratch --srpm`
+   - If `rhpkg build --scratch` fails on a private branch, fall back to the two-step approach: `rhpkg srpm` then `brew build --scratch <branch>-candidate *.src.rpm`
```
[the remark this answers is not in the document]
saschagrunert: Fixed

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,261 @@
...
+
+4. **Declare new sources:**
+   ```bash
+   rhpkg new-sources <package>-*.tar.gz
+   ```
+
+5. **Reset Release and bump the changelog:**
+   First, reset Release in the spec file for the new Version:
+   ```bash
+   sed -i 's/^Release:.*/Release: 0%{?dist}/' <package>.spec
```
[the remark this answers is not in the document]
saschagrunert: Fixed

saschagrunert:
@harche @mrunalp @haircommander @rphillips ready for review

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,272 @@
...
+   ```
+   If any tool is missing, print installation instructions and stop.
+
+   **With `--vagrant`:**
+   - Verify `vagrant` is installed locally.
+   - Verify libvirt is available (`virsh --version`).
+   - Create the work directory: `mkdir -p .work/node-rpm`
+   - If `.work/node-rpm/Vagrantfile` does not exist, copy it from the vendored reference:
+     ```bash
+     cp plugins/node-rpm/references/Vagrantfile .work/node-rpm/Vagrantfile
```
harche: This path only resolves from a checkout of this repo. Once the plugin is installed via `/plugin install node-rpm@ai-helpers` the files live in the plugin cache, so this `cp` fails and the `--vagrant` flow never starts. Other plugins use `${CLAUDE_PLUGIN_ROOT}` for bundled assets, e.g. https://github.com/openshift-eng/ai-helpers/blob/main/plugins/ci/commands/fetch-payloads.md#L26

```suggestion
     cp "${CLAUDE_PLUGIN_ROOT}/references/Vagrantfile" .work/node-rpm/Vagrantfile
```

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,272 @@
...
+   - For cri-tools: branches follow `rhaos-<version>-rhel-{8,9,10}`. Most OCP versions have two RHEL branches (e.g. RHEL 9 + RHEL 10 for OCP 5.0). Check which branches exist for the target version and repeat Phases 1-3 for each branch.
+
+---
+
+### Execution Mode
+
+All shell commands in Phases 1-3 run differently depending on `--vagrant`:
+
+- **Local (default):** Commands run directly in the shell.
+- **Vagrant:** Commands are prefixed with `cd .work/node-rpm && vagrant ssh -c "..."`. The dist-git clone and build happen inside the VM. Output is captured and shown to the user as normal.
```
harche: Each `vagrant ssh -c` opens a fresh session in `/home/vagrant`, so the `cd <package>` from Phase 1 (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L128) does not carry over to later invocations. The confirmation pauses in Phase 2 step 6 and Phase 3 guarantee separate sessions, so `spectool` (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L162), the Release reset (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L173) and `git commit` / `git push` / `rhpkg build` (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L190-L203) would run outside the clone. Worth spelling out that the prefix for everything after the clone should be `vagrant ssh -c "cd <package> && ..."`.

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,272 @@
...
+     ```bash
+     cd .work/node-rpm && vagrant ssh
+     # inside the VM:
+     git config --global user.name "Your Name"
+     git config --global user.email "you@redhat.com"
+     exit
+     ```
+
+5. **Resolve the dist-git branch:**
+   - Read the version mapping from [`../../node-team/skills/node/references/shared/version-map.md`](../../node-team/skills/node/references/shared/version-map.md) to determine the release branch.
```
harche: version-map.md only maps OCP versions to upstream GitHub branches (`release-1.X`, `release-4.Y`/`release-5.Y`, see https://github.com/openshift-eng/ai-helpers/blob/main/plugins/node-team/skills/node/references/shared/version-map.md#L24). It has no `rhaos-*` dist-git data, and the mapping it does have cannot derive one (OCP 5.0 maps to K8s 1.36, but the branch is `rhaos-5.0-rhel-*`). The branch actually comes from `--ocp-version` plus the table in rpm-workflow.md, so I would point this step there and drop the version-map reference.

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,272 @@
...
+1. **Create work directory** (local mode only; Vagrant mode already created this in Phase 0):
+   ```bash
+   mkdir -p .work/node-rpm
+   ```
+
+2. **Clone the dist-git repo:**
+   ```bash
+   cd .work/node-rpm
+   ```
+   If `.work/node-rpm/<package>` already exists (from a prior interrupted run), offer to reuse it (`cd <package> && git fetch && git checkout <branch> && git reset --hard origin/<branch>`) or remove it (`rm -rf <package>`) before proceeding.
```
harche: This checks the host path, but with `--vagrant` the clone lives inside the VM (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L131). A rerun after an interruption passes this check and then fails on `rhpkg clone` hitting the existing directory in the VM. In vagrant mode the check needs to run inside the VM, e.g. `vagrant ssh -c 'test -d <package>'`.

plugins/node-rpm/commands/bump.md
```diff
@@ -0,0 +1,272 @@
...
+   - `<new-version>` (required): The target upstream version, e.g. `1.36.0`
+
+2. **Parse optional arguments:**
+   - `--ocp-version <version>`: Target OCP version. If omitted, prompt the user to specify one.
+   - `--scratch`: Run a scratch build instead of a full build.
+   - `--vagrant`: Run the workflow inside a Vagrant VM.
+
+3. **Check VPN connectivity:**
+   ```bash
+   curl -s --connect-timeout 5 http://download.devel.redhat.com > /dev/null 2>&1 || echo "UNREACHABLE"
```
harche, line 37: This runs on the host, but with `--vagrant` the actual `rhpkg` traffic originates inside the libvirt guest. NAT usually inherits host routes so it mostly works, but with split tunnel or DNS quirks the precheck can pass while the VM cannot reach `pkgs.devel.redhat.com`. Cheap insurance would be rerunning the same curl inside the VM once it is up.

docs/index.html
```diff
@@ -1956,6 +1956,38 @@ <h1>ai-helpers</h1>
         "name": "github.com/openshift-eng"
       }
     },
+    {
```
harche, line 1959: CONTRIBUTING.md says this file is auto generated and should not be edited manually (https://github.com/openshift-eng/ai-helpers/blob/main/CONTRIBUTING.md#L64-L69). `docs/data.json` from the same generator was not updated here, so the next `make update` will produce a surprise diff. I would drop the hand edit and run `make update` instead.

plugins/node-rpm/references/rpm-workflow.md
```diff
@@ -0,0 +1,172 @@
...
+| Package | Upstream Org | Upstream Repo | Dist-git Branch Pattern | Spec Conventions |
+|---------|-------------|--------------|------------------------|------------------|
+| cri-tools | kubernetes-sigs | cri-tools | `rhaos-<version>-rhel-{8,9,10}` (most OCP versions have two RHEL branches; check which exist for the target version) | `commit0`, `Version`, `Release` |
+
+> **Note:** Only cri-tools is currently supported. To add a new package, add a
+> row to this table with its upstream org, repo, branch pattern, and spec
+> conventions, then the `/node-rpm:bump` command will pick it up automatically.
+> If the package uses a non-standard Release format (e.g.
+> `1.git%{shortcommit0}%{?dist}`), document it in the Spec Conventions column
+> so the Release reset step can adapt.
```
harche: This note promises the Release reset step can adapt via the Spec Conventions column, but nothing in bump.md reads that column: the reset is a hardcoded sed (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L173), and `%global commit0` (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L155), the `v` tag prefix (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L148), the tarball glob (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L161), the cri-tools branch clause (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L100) and the Phase 4 next steps (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/commands/bump.md#L220-L221) are equally baked in. The Upstream Org/Repo columns also duplicate components.md, which says other plugins should reference it instead of keeping copies (https://github.com/openshift-eng/ai-helpers/blob/main/plugins/node-team/skills/node/references/shared/components.md#L3-L5). Fine for a cri-tools only v0.0.1, but I would soften this note (or actually wire the command to the table) so adding a package does not silently get its Release scheme rewritten.

plugins/node-rpm/references/rpm-workflow.md
```diff
@@ -0,0 +1,172 @@
...
+6. Update cri-tools in `cri-o/packaging`
+7. Bump the downstream RPM for OpenShift (this is the `/node-rpm:bump` workflow)
+
+## Prerequisites
+
+- **VPN**: Connected to the Red Hat VPN
+- **Kerberos auth**: `kinit user@IPA.REDHAT.COM` (verify with `klist`)
+- **Tools installed**: `rhpkg`, `spectool`, `rpmdev-bumpspec`
+
+## Environment Setup
```
harche, line 98: This section duplicates what the Vagrantfile provisioning already does, and the two ssh configs have already diverged: the Vagrantfile copy has `StrictHostKeyChecking` and `ServerAlive*` globals plus the `*.iad2` and `*.rdu2` bastion blocks (https://github.com/openshift-eng/ai-helpers/blob/4cbd6596becc0827c83e21733d2d85a6dcb2c0ef/plugins/node-rpm/references/Vagrantfile#L49-L80) that this one drops. Same story for Workflow Steps above, which repeats bump.md Phases 1 to 3 command for command. A bare metal user following this doc gets a different config than the VM uses, and future fixes will land in one copy only. I would keep the rationale here and point at the Vagrantfile (and the command) as the source of truth.

harche (commented):
Thanks for this PR! Left a few comments inline, mostly around the `--vagrant` flow.

plugins/node-rpm/commands/bump.md
saschagrunert: Fixed. Using `${CLAUDE_PLUGIN_ROOT}/references/Vagrantfile` now.

plugins/node-rpm/commands/bump.md
saschagrunert: Updated the Execution Mode section to document that each `vagrant ssh -c` opens a fresh session in `/home/vagrant`, so all commands after the clone need `cd <package> && ...` prefix.

plugins/node-rpm/commands/bump.md
saschagrunert: Dropped the version-map.md reference. Branch resolution now points only to rpm-workflow.md's Dist-git Branch Pattern column plus `--ocp-version`.

plugins/node-rpm/commands/bump.md
saschagrunert: In vagrant mode the check now runs inside the VM: `vagrant ssh -c "test -d <package>"`.

plugins/node-rpm/commands/bump.md
saschagrunert: Added a VPN connectivity recheck inside the VM after startup to catch split-tunnel or DNS issues.

docs/index.html
saschagrunert: Reverted hand edit and regenerated with `make update`.

plugins/node-rpm/references/rpm-workflow.md
saschagrunert: Softened the note. It now explicitly states that bump.md hardcodes the cri-tools workflow and adding a package with different conventions requires updating bump.md to read the Spec Conventions column.

plugins/node-rpm/references/rpm-workflow.md
saschagrunert: Replaced the duplicated setup steps and SSH config with a pointer to the Vagrantfile as the source of truth. Bare-metal users are directed to read the Vagrantfile provisioning script.

PR #594, opened 2026-07-01
node-bug: add bug triage and assignment plugin

saschagrunert:
Adds a `node-bug` plugin for Node team bug triage. Queries open bugs from OCPBUGS using the "Node Bugs" saved filter (ID 83963), classifies into triage buckets (Release Blockers, Potential Blockers, Customer Escalations, Component Regressions, Untriaged), routes to sub-teams (Core/Devices/Kueue), suggests assignments based on workload, and generates triage summaries. Optionally posts to Slack.
Part of the [node-team plugin family](../node-team/).

- Independent from #593 (node-rpm)
- Read-only: no Jira writes, all suggestions are advisory
- Requires `JIRA_API_TOKEN` for Jira access
- Optional team roster files at `~/.node-assistant/team-roster-{core,dra}.json` for assignment suggestions
- References the "Node Bugs" dashboard (ID 12991) for visual overview

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a new `node-bug` plugin with the `/node-bug:triage` command, including sub-team/sprint and unassigned-only filtering, severity bucketing, workload-aware assignment suggestions, and saved triage reports (with optional Slack notifications).
* **Bug Fixes**
  * Updated the Node team workflow so bug triage/assignment route to the new `node-bug` flow.
* **Documentation**
  * Published plugin and command documentation; updated marketplace entries and corrected Node component/sub-team and JIRA guidance.
* **Chores**
  * Updated plugin versioning and clarified plugin ownership/reviewers.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1867,6 +1867,38 @@ footer a { color: var(--primary); }
       }
     },
     {
+      "name": "node-bug",
+      "description": "Node-specific bug triage, sub-team routing, and assignment suggestions for OpenShift Node team components",
+      "version": "0.0.1",
+      "has_readme": true,
+      "commands": [
+        {
+          "name": "triage",
+          "full_name": "node-bug:triage",
+          "description": "Query open Node bugs, classify by severity and sub-team, suggest assignments, and generate a triage summary",
+          "description_html": "Query open Node bugs, classify by severity and sub-team, suggest assignments, and generate a triage summary",
+          "synopsis": "/node-bug:triage [--sub-team core|devices|kueue] [--sprint \"OCP Node Core Sprint 42\"] [--unassigned-only]",
+          "body_html": "\u003cp\u003eQueries all open bugs in OCPBUGS for Node team components using the &quot;Node Bugs&quot; saved filter, classifies them into triage buckets (Release Blockers, Customer Escalations, Potential Blockers, Component Regressions, Untriaged), routes each bug to the correct sub-team, and suggests assignments based on current workload.\u003c/p\u003e\u003cp\u003eDesigned for both interactive triage sessions and headless execution via \u003ccode\u003eclaude --print\u003c/code\u003e.\u003c/p\u003e"
+        }
+      ],
+      "skills": [],
+      "agents": [],
+      "hooks": [],
+      "mcp_servers": [],
+      "rules": [],
+      "category": "openshift",
+      "keywords": [
+        "openshift",
+        "node",
+        "bugs",
+        "triage",
+        "jira"
+      ],
+      "author": {
+        "name": "github.com/openshift-eng"
+      }
+    },
+    {
       "name": "node-cve",
       "description": "CVE triage for OpenShift Node team components",
       "version": "0.0.4",
@@ -1991,7 +2023,7 @@ footer a { color: var(--primary); }
     {
       "name": "node-team",
       "description": "OpenShift Node team assistant for development, deployment, debugging, and workflow tasks across kubelet, MCO, CRI-O, crun, conmonrs, Kueue operator, Jira, Red Hat KB/support cases, Prometheus, and platform docs.",
-      "version": "0.17.1",
+      "version": "0.18.0",
       "has_readme": true,
       "commands": [
         {
@@ -2030,8 +2062,8 @@ footer a { color: var(--primary); }
       "skills": [
         {
           "name": "node",
-          "description": "OpenShift Node team assistant. Covers kubelet, MCO, CRI-O, crun, conmonrs, Kueue operator, Jira (OCPNODE/OCPBUGS), Red Hat KB/support cases, Prometheus, and K8s/OCP docs. Triggers on OpenShift node-layer development, deployment, debugging, or team workflow tasks. For CVE/vulnerability triage, analysis, or reporting, defer to the node-cve plugin.",
-          "description_html": "OpenShift Node team assistant. Covers kubelet, MCO, CRI-O, crun, conmonrs, Kueue operator, Jira (OCPNODE/OCPBUGS), Red Hat KB/support cases, Prometheus, and K8s/OCP docs. Triggers on OpenShift node-layer development, deployment, debugging, or team workflow tasks. For CVE/vulnerability triage, analysis, or reporting, defer to the node-cve plugin.",
+          "description": "OpenShift Node team assistant. Covers kubelet, MCO, CRI-O, crun, conmonrs, Kueue operator, Jira (OCPNODE/OCPBUGS), Red Hat KB/support cases, Prometheus, and K8s/OCP docs. Triggers on OpenShift node-layer development, deployment, debugging, or team workflow tasks. For CVE/vulnerability triage, analysis, or reporting, defer to node-cve. For bug triage and assignment, defer to node-bug.",
+          "description_html": "OpenShift Node team assistant. Covers kubelet, MCO, CRI-O, crun, conmonrs, Kueue operator, Jira (OCPNODE/OCPBUGS), Red Hat KB/support cases, Prometheus, and K8s/OCP docs. Triggers on OpenShift node-layer development, deployment, debugging, or team workflow tasks. For CVE/vulnerability triage, analysis, or reporting, defer to node-cve. For bug triage and assignment, defer to node-bug.",
           "meta": "Tools: Bash(curl:*)"
         }
       ],
--- a/plugins/node-bug/commands/triage.md
+++ b/plugins/node-bug/commands/triage.md
@@ -1,0 +1,224 @@
+---
+description: Query open Node bugs, classify by severity and sub-team, suggest assignments, and generate a triage summary
+argument-hint: "[--sub-team core|devices|kueue] [--sprint <name>] [--unassigned-only]"
+---
+
+## Name
+node-bug:triage
+
+## Synopsis
+```text
+/node-bug:triage [--sub-team core|devices|kueue] [--sprint "OCP Node Core Sprint 42"] [--unassigned-only]
+```
+
+## Description
+
+Queries all open bugs in OCPBUGS for Node team components using the "Node Bugs" saved filter, classifies them into triage buckets (Release Blockers, Customer Escalations, Potential Blockers, Component Regressions, Untriaged), routes each bug to the correct sub-team, and suggests assignments based on current workload.
+
+Designed for both interactive triage sessions and headless execution via `claude --print`.
+
+## Implementation
+
+### Phase 0: Setup and Argument Parsing
+
+1. **Parse Arguments**
+   - `--sub-team core|devices|kueue`: Filter results to one sub-team's components. Optional. Read sub-team component lists from the sub-teams table in [shared/components.md](../../node-team/skills/node/references/shared/components.md) rather than hardcoding names.
+     - `core`: all Node components not listed under another sub-team
+     - `devices`: components listed under DRA/Devices in the sub-teams table
+     - `kueue`: components listed under Kueue in the sub-teams table
+   - `--sprint <name>`: Filter to bugs in a specific sprint (e.g., "OCP Node Core Sprint 42"). Optional.
+   - `--unassigned-only`: Show only untriaged or unassigned bugs. Optional.
+2. **Validate Jira Credentials**
+
+   Use the authentication chain from the [jira reference](../../node-team/skills/node/references/jira.md):
+
+   ```bash
+   JIRA_API_TOKEN="${JIRA_API_TOKEN:-$(security find-generic-password -s "JIRA_API_TOKEN" -w 2>/dev/null || secret-tool lookup service redhat key JIRA_API_TOKEN 2>/dev/null)}"
+   JIRA_USER="${JIRA_EMAIL:-$(security find-generic-password -s "JIRA_API_TOKEN" -g 2>&1 | grep acct | sed 's/.*="//;s/"//')}"
+   : "${JIRA_USER:=$(git config user.email)}"
+   [[ "$JIRA_USER" != *@* ]] && JIRA_USER="${JIRA_USER}@redhat.com"
+   ```
+
+   Exit with error if `JIRA_API_TOKEN` is empty after the chain.
+
+3. **Create work directory**: `mkdir -p .work/node-bug/triage-$(date +%Y-%m-%d)`
+
+---
+
+### Phase 1: Query Bugs
+
+1. **Use the Jira saved filter** "Node Bugs" (ID 83963) for the base query. The filter defines which components are in scope.
+
+2. **Build JQL**:
+
+   ```text
+   filter = "Node Bugs" AND status not in (Closed, Done, Verified)
+   ```
+
+   Apply optional filters (read sub-team component names from the sub-teams table in [shared/components.md](../../node-team/skills/node/references/shared/components.md)):
+   - If `--sub-team devices`: append `AND component in (<DRA/Devices components>)`
+   - If `--sub-team kueue`: append `AND component in (<Kueue components>)`
+   - If `--sub-team core`: append `AND component not in (<DRA/Devices components>, <Kueue components>)`
+   - If `--sprint <name>`: append `AND sprint = "<name>"`
+   - If `--unassigned-only`: append `AND (assignee is EMPTY OR assignee in ("aos-node@redhat.com") OR priority = Undefined OR "Release Blocker" = Proposed)`
+
+3. **Execute the query** using the POST search endpoint:
+
+   ```bash
+   curl -s -u "$JIRA_USER:$JIRA_API_TOKEN" \
+     -H "Content-Type: application/json" \
+     -X POST "https://redhat.atlassian.net/rest/api/3/search/jql" \
+     -d '{
+       "jql": "<constructed JQL>",
+       "maxResults": 100,
+       "fields": [
+         "key", "summary", "status", "priority", "assignee",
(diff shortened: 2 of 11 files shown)


saschagrunert:
@harche @mrunalp @haircommander @rphillips ready for review

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+Queries all open bugs in OCPBUGS for Node team components using the "Node Bugs" saved filter, classifies them into triage buckets (Release Blockers, Customer Escalations, Potential Blockers, Component Regressions, Untriaged), routes each bug to the correct sub-team, and suggests assignments based on current workload. Optionally sends a triage summary to Slack.
+
+Designed for both interactive triage sessions and headless execution via `claude --print`.
+
+## Implementation
+
+### Phase 0: Setup and Argument Parsing
+
+1. **Parse Arguments**
+   - `--sub-team core|dra|kueue`: Filter results to one sub-team's components. Optional. Read sub-team component lists from the sub-teams table in [shared/components.md](../../node-team/skills/node/references/shared/components.md) rather than hardcoding names.
```
haircommander: nit: devices not dra

haircommander:
one nit, otherwise LGTM

plugins/node-bug/commands/triage.md
saschagrunert: Fixed :+1:

plugins/node-team/skills/node/references/jira.md
```diff
@@ -119,7 +119,7 @@ Bot account treated as unassigned: `Node Team Bot Account`.
 | Team | Sprint filter | Roster file | Bug components |
 |------|--------------|-------------|----------------|
 | Core | `Node Core` | `team-roster-core.json` | All Node components |
-| DRA/Devices | `Node Devices` | `team-roster-dra.json` | Node / Device Manager, Node / Instaslice-operator |
+| DRA/Devices | `Node Devices` | `team-roster-dra.json` | Node / Device Manage, Node / Instaslice-operator |
```
harche: This drops the trailing r from the real component name. I checked the live OCPBUGS component list (`curl -s "https://redhat.atlassian.net/rest/api/3/project/OCPBUGS/component?query=Device"`, works anonymously) and it returns `Node / Device Manager`. Since `/node-bug:triage` builds `component in (...)` from this exact string, the typo makes `--sub-team devices` silently match zero Device Manager bugs, and `--sub-team core` picks them up instead via `component not in (...)`. The same typo already exists in components.md (pre existing on the unchanged DRA row), so it is worth fixing there too rather than aligning jira.md to it. For the record, `Node / Kueue` and `Node / Instaslice-operator` are correct as written.

```suggestion
| DRA/Devices | `Node Devices` | `team-roster-dra.json` | Node / Device Manager, Node / Instaslice-operator |
```

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+     }'
+   ```
+
+   Custom field mapping:
+   - `customfield_10840`: Severity
+   - `customfield_10847`: Release Blocker
+   - `customfield_10978`: SFDC Cases Counter
+   - `customfield_10020`: Sprint
+
+   Handle pagination if `total` exceeds `maxResults` by issuing additional requests with `startAt`.
```
harche: This pagination can never trigger: I tested `POST /rest/api/3/search/jql` with a real token and the response contains only `issues`, `isLast`, and `nextPageToken`. There is no `total` field, and this endpoint ignores `startAt` (see https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-jql-post). With more than 100 open bugs the run silently drops everything past the first page while looking complete.

```suggestion
   Handle pagination via `nextPageToken`: while the response has `isLast: false`, repeat the request with `"nextPageToken": "<token from previous response>"` in the body. This endpoint does not return `total` and does not accept `startAt`.
```

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+
+2. **Validate Jira Credentials**
+
+   Use the authentication chain from the [jira reference](../../node-team/skills/node/references/jira.md):
+
+   ```bash
+   JIRA_API_TOKEN="${JIRA_API_TOKEN:-$(security find-generic-password -s "JIRA_API_TOKEN" -w 2>/dev/null || secret-tool lookup service redhat key JIRA_API_TOKEN 2>/dev/null)}"
+   JIRA_USER="${JIRA_EMAIL:-$(security find-generic-password -s "JIRA_API_TOKEN" -g 2>&1 | grep acct | sed 's/.*="//;s/"//')}"
+   [[ "$JIRA_USER" != *@* ]] && JIRA_USER="${JIRA_USER}@redhat.com"
+   : "${JIRA_USER:=$(git config user.email)}"
```
harche: These two lines are in the wrong order. When the keychain lookup yields nothing (always on Linux, where `security` does not exist, with `JIRA_EMAIL` unset), `JIRA_USER` is empty, the `!= *@*` test matches, and `JIRA_USER` becomes the literal `@redhat.com`. The `git config user.email` fallback on the next line is then dead code because the variable is already non empty, and every curl authenticates as `-u "@redhat.com:token"` and gets a 401. The snippet is copied verbatim from the jira reference (https://github.com/openshift-eng/ai-helpers/blob/main/plugins/node-team/skills/node/references/jira.md), so the fix belongs there as well.

```suggestion
   : "${JIRA_USER:=$(git config user.email)}"
   [[ "$JIRA_USER" != *@* ]] && JIRA_USER="${JIRA_USER}@redhat.com"
```

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+   ```text
+   filter = "Node Bugs" AND status not in (Closed, Done, Verified)
+   ```
+
+   Apply optional filters (read sub-team component names from the sub-teams table in [shared/components.md](../../node-team/skills/node/references/shared/components.md)):
+   - If `--sub-team devices`: append `AND component in (<DRA/Devices components>)`
+   - If `--sub-team kueue`: append `AND component in (<Kueue components>)`
+   - If `--sub-team core`: append `AND component not in (<DRA/Devices components>, <Kueue components>)`
+   - If `--sprint <name>`: append `AND sprint = "<name>"`
+   - If `--unassigned-only`: append `AND (assignee is EMPTY OR assignee in ("aos-node@redhat.com") OR priority = Undefined)`
```
harche: This filter misses one leg of your own Untriaged definition. The bucket at https://github.com/openshift-eng/ai-helpers/blob/b67a18bc062d78bd2adfaf5fa163513d3ba04857/plugins/node-bug/commands/triage.md#L118 counts a bug as untriaged when `"Release Blocker"` is `Proposed`, matching the canonical definition in jira.md, but this JQL omits that clause. A bug with a real assignee and a set priority whose release blocker decision is still Proposed is excluded from `--unassigned-only` even though the flag is documented as showing untriaged bugs.

```suggestion
   - If `--unassigned-only`: append `AND (assignee is EMPTY OR assignee in ("aos-node@redhat.com") OR priority = Undefined OR "Release Blocker" = Proposed)`
```

plugins/node-team/skills/node/references/jira.md
```diff
@@ -119,7 +119,7 @@ Bot account treated as unassigned: `Node Team Bot Account`.
 | Team | Sprint filter | Roster file | Bug components |
 |------|--------------|-------------|----------------|
 | Core | `Node Core` | `team-roster-core.json` | All Node components |
```
harche: This table and the one in shared/components.md were mirror copies, and the PR now updates only the components.md copy (https://github.com/openshift-eng/ai-helpers/blob/b67a18bc062d78bd2adfaf5fa163513d3ba04857/plugins/node-team/skills/node/references/shared/components.md#L76-L80 gets the Kueue row and the new Core wording). This copy still says Core owns `All Node components` and has no Kueue row, so the two tables disagree about who owns `Node / Kueue`. Suggest either applying the same edits here or collapsing to one canonical table (components.md already declares itself the canonical source that other plugins should reference) and linking to it from here.

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+   - **Customer Escalations**: `customfield_10978` (SFDC Cases Counter) is not null/empty. To also catch bugs marked only via Customer Impact (no SFDC case linked), run a supplementary JQL query: `filter = "Node Bugs" AND "Customer Impact" = "Customer Escalated" AND "SFDC Cases Counter" is EMPTY` and merge the results. The Customer Impact custom field ID is not documented, so JQL is the only way to check it.
+   - **Component Regressions**: labels contain `component-regression`
+   - **Untriaged**: priority is "Undefined", OR `"Release Blocker"` is "Proposed", OR assignee is `aos-node@redhat.com`
+   - **Other**: all remaining bugs
+
+   A bug can appear in multiple buckets (e.g., a release blocker that is also a customer escalation). Count it in each applicable bucket.
+
+3. **Assignment suggestions** (when team roster files exist):
+
+   Load team rosters from `~/.node-assistant/team-roster-core.json` and `~/.node-assistant/team-roster-dra.json`. Kueue members are tracked under the Core roster. If the files do not exist, skip assignment suggestions and print "Roster files not found, skipping assignment suggestions."
```
harche: The Kueue assignment path cannot work as specified. Roster files are flat display name to github handle maps per the Team Roster section in jira.md, with no sub team marker, and Kueue members sit inside `team-roster-core.json` indistinguishable from Core members. So "suggest the team member with the fewest open bugs from the appropriate sub-team roster" (https://github.com/openshift-eng/ai-helpers/blob/b67a18bc062d78bd2adfaf5fa163513d3ba04857/plugins/node-bug/commands/triage.md#L135) draws from the entire Core roster for a Kueue bug and can suggest someone who does not work on Kueue. A `team-roster-kueue.json` (or an explicit member list in the roster format) would make the suggestion meaningful.

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+   ```bash
+   JIRA_API_TOKEN="${JIRA_API_TOKEN:-$(security find-generic-password -s "JIRA_API_TOKEN" -w 2>/dev/null || secret-tool lookup service redhat key JIRA_API_TOKEN 2>/dev/null)}"
+   JIRA_USER="${JIRA_EMAIL:-$(security find-generic-password -s "JIRA_API_TOKEN" -g 2>&1 | grep acct | sed 's/.*="//;s/"//')}"
+   [[ "$JIRA_USER" != *@* ]] && JIRA_USER="${JIRA_USER}@redhat.com"
+   : "${JIRA_USER:=$(git config user.email)}"
+   ```
+
+   Exit with error if `JIRA_API_TOKEN` is empty after the chain.
+
+3. **If `--notify-slack`**, verify Slack credentials. Either `$SLACK_API_TOKEN` + `$SLACK_CHANNEL` (preferred) or `$SLACK_WEBHOOK`. Exit with error if neither is configured. Default `$SLACK_CHANNEL` to `GK6BJJ1J5` (`#team-node`) if not set.
```
harche: Two issues here. First, the spec contradicts itself: it requires `$SLACK_API_TOKEN` plus `$SLACK_CHANNEL` (or the webhook) and says to exit if neither is configured, but the next sentence defaults `$SLACK_CHANNEL` unconditionally, so with only a token set the executing model can justify either aborting or posting to the hardcoded channel. Second, and more fundamentally: as far as I know the Node team has no Slack app or API token access under current security policy, and incoming webhooks are also app scoped, so both paths may be unprovisionable for the intended users. Where would a working token or webhook come from? If there is no sanctioned source, consider dropping `--notify-slack` for now instead of shipping a flag nobody can use. (Minor: the claim that the token path "enables threading" is also not realized, since the `chat.postMessage` call at https://github.com/openshift-eng/ai-helpers/blob/b67a18bc062d78bd2adfaf5fa163513d3ba04857/plugins/node-bug/commands/triage.md#L191-L194 never sends `thread_ts`.)

plugins/node-bug/commands/triage.md
```diff
@@ -0,0 +1,253 @@
...
+1. **Route each bug to its sub-team** using the sub-teams table from [shared/components.md](../../node-team/skills/node/references/shared/components.md):
+   - DRA/Devices: bugs whose component appears in the DRA/Devices row of the sub-teams table
+   - Kueue: bugs whose component appears in the Kueue row of the sub-teams table
+   - Core: all remaining Node components
+
+2. **Classify each bug into triage buckets** using the [Bug Triage Definitions](../../node-team/skills/node/references/jira.md):
+
+   - **Release Blockers**: `"Release Blocker"` field value is "Approved"
+   - **Potential Blockers**: priority is "Blocker" AND `"Release Blocker"` is empty (Blocker priority set but no release blocker decision yet)
+   - **Customer Escalations**: `customfield_10978` (SFDC Cases Counter) is not null/empty. To also catch bugs marked only via Customer Impact (no SFDC case linked), run a supplementary JQL query: `filter = "Node Bugs" AND "Customer Impact" = "Customer Escalated" AND "SFDC Cases Counter" is EMPTY` and merge the results. The Customer Impact custom field ID is not documented, so JQL is the only way to check it.
```
harche: The premise here is not quite right: the Customer Impact field ID is discoverable. `curl -s "https://redhat.atlassian.net/rest/api/3/field"` (anonymous) returns it as `customfield_10689`. Adding that row to the Custom Field IDs table in jira.md (this PR already edits that file) and requesting `customfield_10689` in the Phase 1 fields array lets this bucket classify in code like the others, eliminating the supplementary query, its merge logic, and the snapshot skew between two queries against a live filter.

docs/index.html
```diff
@@ -1866,6 +1866,38 @@ <h1>ai-helpers</h1>
         "name": "github.com/openshift-eng"
       }
     },
+    {
```
harche, line 1869: Same note as on #593: CONTRIBUTING.md marks this file as auto generated and not to be edited manually (https://github.com/openshift-eng/ai-helpers/blob/main/CONTRIBUTING.md#L64-L69), and `docs/data.json` at this PR head does not contain node-bug, which shows `make update` was not run. Suggest dropping the hand edit and regenerating.

harche (commented):
Thanks for this PR! Left some comments inline. A few of them were checked directly against the live Jira API.

plugins/node-team/skills/node/references/jira.md
saschagrunert: Fixed to `Device Manager` in jira.md and all occurrences in components.md.

plugins/node-bug/commands/triage.md
saschagrunert: Replaced with `nextPageToken`/`isLast` pagination per the v3 search endpoint.

plugins/node-bug/commands/triage.md
saschagrunert: Swapped the lines in both triage.md and jira.md so `git config user.email` fallback runs before the `@redhat.com` append.

plugins/node-bug/commands/triage.md
saschagrunert: Added `OR "Release Blocker" = Proposed` to the `--unassigned-only` JQL clause and updated the argument description.

plugins/node-team/skills/node/references/jira.md
saschagrunert: Synced: added Kueue row, updated Core to "All Node components not listed under another sub-team", fixed Device Manager. Both tables now match.

plugins/node-bug/commands/triage.md
saschagrunert: Added `team-roster-kueue.json` as a separate roster file for the Kueue sub-team in jira.md, components.md, and triage.md.

plugins/node-bug/commands/triage.md
saschagrunert: Dropped `--notify-slack` entirely: removed the flag, Slack credentials check, posting code, and related examples.

plugins/node-bug/commands/triage.md
saschagrunert: Added `customfield_10689` (Customer Impact) to the jira.md Custom Field IDs table and the Phase 1 fields array. Replaced the supplementary JQL query with a direct field check.

docs/index.html
saschagrunert: Reverted hand edit and regenerated with `make update`.

plugins/node-team/skills/node/references/jira.md
```diff
@@ -118,8 +118,9 @@ Bot account treated as unassigned: `Node Team Bot Account`.
 
 | Team | Sprint filter | Roster file | Bug components |
 |------|--------------|-------------|----------------|
-| Core | `Node Core` | `team-roster-core.json` | All Node components |
+| Core | `Node Core` | `team-roster-core.json` | All Node components not listed under another sub-team |
 | DRA/Devices | `Node Devices` | `team-roster-dra.json` | Node / Device Manager, Node / Instaslice-operator |
+| Kueue | `OCP Kueue` | `team-roster-kueue.json` | Node / Kueue |
```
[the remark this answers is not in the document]
saschagrunert: See https://github.com/openshift-eng/ai-helpers/pull/594#discussion_r3515535704

saschagrunert:
@harche had to rebase

PR #595, opened 2026-07-01
node-team: enhance overview command with onboarding content

saschagrunert:
## What this PR does / why we need it:
Rewrites the `node-team:overview` command to be a comprehensive entry point for the Node team. Incorporates content from the Node Team Onboarding Guide and existing shared references.

New sections: team mission, responsibilities, component ownership, sub-team structure (Core, DRA/Devices, Kueue), ceremonies with sprint cadence, upstream SIG-Node community, Slack channels and user groups, mailing lists, customer support tools (SupportShell, yank, omc), plugin routing table (all 8 commands across 5 plugins), and key links.

Also adds a Kueue row to the sub-teams tables in `components.md` and `jira.md`, fixes the "Device Manager" -> "Device Manage" typo in `jira.md`, and renames the `--sub-team` flag value from `dra` to `devices` to match the DRA/Devices sub-team name.

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [x] This change includes docs.

## Summary by CodeRabbit

* **New Features**
  * Updated the Node team overview guidance to cover more team areas and include additional sub-team options.
  * Added support for the new **Kueue** sub-team in the Node team reference materials.

* **Documentation**
  * Refreshed team reference content with updated responsibilities, links, meeting info, and support resources.
  * Corrected several Node component names and mapping details for clearer navigation.

* **Bug Fixes**
  * Fixed outdated sub-team and component references so the latest team structure is shown consistently.

PR #584, opened 2026-06-29
node-team: add medium-risk compliance controls and hierarchy

saschagrunert:
## What this PR does / why we need it:
Formalizes node-team as the umbrella plugin with node-cve as a dependency,
and adds compliance documentation required by Red Hat's medium-risk AI agent
controls.

Changes:
- Hierarchy: node-cve declares node-team as a dependency in plugin.json
- TR-01/HU-01: AI-generated attribution labels on Jira comments and Slack messages
- TR-02: User guide (`docs/compliance/user-guide.md`)
- GLC-08: Capabilities inventory (`docs/compliance/capabilities-inventory.md`)
- [DATA-02](https://redhat.atlassian.net/browse/DATA-02): Dataflow diagram (`docs/compliance/dataflow.md`)
- Mon-01: Monitoring plan (`docs/compliance/monitoring-plan.md`)
- [DATA-05](https://redhat.atlassian.net/browse/DATA-05): `/node-team:cleanup` command for artifact purge
- Plugin family table in README with planned satellites (bug-triage, onboarding, rpm)
- Version bumps: node-cve 0.0.4, node-team 0.17.0

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [x] This change includes docs.

PR #603, opened 2026-07-06
node-team: fix consistency across plugin family

saschagrunert:
Fixes consistency issues across the node-team plugin family after PRs #584, #589, #593, #594, and #595 merged.

**Versioning:**
- Bump node-team to v1.0.0; all satellites depend on `^1.0.0` (the old `^0.17.0` excluded 0.18.x under semver 0.x rules)

**node-bug:**
- Fix triage bucket definitions to match jira.md Bug Triage Definitions table (Blocker+, Blocker?)
- Fix descriptions ("severity" to "priority"), fix roster list to include kueue
- Keep curl-based Jira access (jira CLI silently drops COMPONENT column and injects default project)

**node-team:**
- Cleanup: add .work/node-bug/ and .work/node-rpm/, fix `vagrant destroy -f` to use subshell (prevents orphaning libvirt domains on failure), update description
- Preflight: annotate jira CLI check as node-cve only, mention node-bug:triage
- Overview: fix roster to include Kueue sub-team
- Compliance: add node-bug and node-rpm to dataflow.md and user-guide.md
- Update jira.md, capabilities inventory, team-info routing table, and README for new satellites

**Satellites:**
- node-cve: add satellite README pattern (Part of, Installation, dependency note)
- node-onboarding: fix "Requires" wording
- node-rpm: standardize "Part of" link format

The node-bug triage command stays on curl rather than switching to jira CLI. The CLI silently skips the COMPONENT column (not in its valid columns list) and injects the configured default project into JQL, both of which break Phase 1 silently. curl's POST search endpoint returns components as structured JSON and executes JQL as-is.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added/updated `node-team:cleanup` documentation to include cached bug triage reports, cloned artifacts, dist-git clones, Vagrant VMs, and roster cache.
  * Expanded installation and quick-start docs for additional node team plugins, including bug triage and RPM workflows.
* **Bug Fixes**
  * Updated bug triage terminology and command rules to use **priority** (including phase bucket logic and rendered output).
  * Refreshed preflight guidance to require Jira checks and to reference the bug triage flow.
* **Documentation / Chores**
  * Updated plugin READMEs and compliance/data-flow docs; removed Slack-related setup/mentions from bug triage instructions.
  * Bumped plugin versions and aligned `node-team` dependency constraints across the suite.

### DIFF
--- a/plugins/node-bug/.claude-plugin/plugin.json
+++ b/plugins/node-bug/.claude-plugin/plugin.json
@@ -1,14 +1,14 @@
 {
   "name": "node-bug",
   "description": "Node-specific bug triage, sub-team routing, and assignment suggestions for OpenShift Node team components",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "author": {
     "name": "github.com/openshift-eng"
   },
   "dependencies": [
     {
       "name": "node-team",
-      "version": "^0.17.0"
+      "version": "^1.0.0"
     }
   ]
 }
--- a/plugins/node-bug/README.md
+++ b/plugins/node-bug/README.md
@@ -1,6 +1,6 @@
 # Node Bug Plugin
 
-Node-specific bug triage for OpenShift Node team components. Queries open bugs from OCPBUGS, classifies by severity and sub-team, suggests assignments based on workload, and generates triage summaries. Complements `/jira:grooming` with Node-specific JQL filters, sub-team routing, and team roster integration.
+Node-specific bug triage for OpenShift Node team components. Queries open bugs from OCPBUGS, classifies by priority and sub-team, suggests assignments based on workload, and generates triage summaries. Complements `/jira:grooming` with Node-specific JQL filters, sub-team routing, and team roster integration.
 
 Part of the [node-team plugin family](../node-team/).
 
@@ -14,9 +14,9 @@ Requires the `node-team` plugin (installed automatically as a dependency).
 
 ## Command
 
-### `/node-bug:triage [--sub-team core|devices|kueue] [--sprint <name>] [--unassigned-only] [--notify-slack]`
+### `/node-bug:triage [--sub-team core|devices|kueue] [--sprint <name>] [--unassigned-only]`
 
-Query open Node bugs, classify by severity and sub-team, suggest assignments, and generate a triage summary.
+Query open Node bugs, classify by priority and sub-team, suggest assignments, and generate a triage summary.
 
 **Examples:**
 
@@ -29,9 +29,6 @@ Query open Node bugs, classify by severity and sub-team, suggest assignments, an
 
 # Show only unassigned bugs for DRA/Devices
 /node-bug:triage --sub-team devices --unassigned-only
-
-# Triage and notify Slack
-/node-bug:triage --notify-slack
 ```
 
 **Arguments:**
@@ -39,11 +36,8 @@ Query open Node bugs, classify by severity and sub-team, suggest assignments, an
 - `--sub-team core|devices|kueue`: Filter to one sub-team's components (Core, DRA/Devices, or Kueue)
 - `--sprint <name>`: Filter to bugs in a specific sprint
 - `--unassigned-only`: Show only untriaged or unassigned bugs
-- `--notify-slack`: Send summary to Slack (requires `SLACK_API_TOKEN` + `SLACK_CHANNEL` or `SLACK_WEBHOOK`)
 
 ## Prerequisites
 
-- `JIRA_API_TOKEN` environment variable (or macOS Keychain / Linux secret-tool)
-- `curl` (for Jira REST API and Slack)
-- Optional: `SLACK_API_TOKEN` + `SLACK_CHANNEL` or `SLACK_WEBHOOK` (for `--notify-slack`)
-- Optional: `~/.node-assistant/team-roster-{core,dra}.json` (for assignment suggestions)
+- `JIRA_API_TOKEN` (env var, macOS Keychain, or Linux secret-tool) and `curl`
+- Optional: `~/.node-assistant/team-roster-{core,dra,kueue}.json` (for assignment suggestions)
--- a/plugins/node-bug/commands/triage.md
+++ b/plugins/node-bug/commands/triage.md
@@ -1,5 +1,5 @@
 ---
-description: Query open Node bugs, classify by severity and sub-team, suggest assignments, and generate a triage summary
+description: Query open Node bugs, classify by priority and sub-team, suggest assignments, and generate a triage summary
 argument-hint: "[--sub-team core|devices|kueue] [--sprint <name>] [--unassigned-only]"
 ---
 
@@ -109,8 +109,8 @@ Designed for both interactive triage sessions and headless execution via `claude
 
 2. **Classify each bug into triage buckets** using the [Bug Triage Definitions](../../node-team/skills/node/references/jira.md):
 
-   - **Release Blockers**: `"Release Blocker"` field value is "Approved"
-   - **Potential Blockers**: priority is "Blocker" AND `"Release Blocker"` is empty (Blocker priority set but no release blocker decision yet)
+   - **Release Blockers**: `"Release Blocker"` field value is "Approved", OR priority is "Blocker"
+   - **Potential Blockers**: `"Release Blocker"` field value is "Proposed", OR (priority is "Blocker" AND `"Release Blocker"` is empty)
    - **Customer Escalations**: `customfield_10978` (SFDC Cases Counter) is not null/empty, OR `customfield_10689` (Customer Impact) value is "Customer Escalated"
    - **Component Regressions**: labels contain `component-regression`
    - **Untriaged**: priority is "Undefined", OR `"Release Blocker"` is "Proposed", OR assignee is `aos-node@redhat.com`
@@ -150,14 +150,14 @@ Designed for both interactive triage sessions and headless execution via `claude
    Other: X
 
    --- Release Blockers ---
-   * OCPBUGS-XXXXX: <summary> (Component, Severity, Assignee)
-   * OCPBUGS-XXXXX: <summary> (Component, Severity, Unassigned -> suggested: <name>)
+   * OCPBUGS-XXXXX: <summary> (Component, Priority, Assignee)
+   * OCPBUGS-XXXXX: <summary> (Component, Priority, Unassigned -> suggested: <name>)
 
    --- Customer Escalations ---
    * OCPBUGS-XXXXX: <summary> (Component, N support cases, Assignee)
 
    --- Potential Blockers ---
-   * OCPBUGS-XXXXX: <summary> (Component, Severity, Assignee)
+   * OCPBUGS-XXXXX: <summary> (Component, Priority, Assignee)
 
    --- Component Regressions ---
    * OCPBUGS-XXXXX: <summary> (Component, Assignee)
--- a/plugins/node-team/commands/cleanup.md
+++ b/plugins/node-team/commands/cleanup.md
@@ -14,8 +14,8 @@ node-team:cleanup
 ## Description
 
 Removes local artifacts created by Node team plugins: triage reports, cloned
-repos, and cached roster data. Use this to reclaim disk space or to satisfy
-data purge requirements (DATA-05).
+repos, dist-git clones, Vagrant VMs, and cached roster data. Use this to
+reclaim disk space or to satisfy data purge requirements (DATA-05).
 
 ## Options
 
@@ -33,16 +33,19 @@ Scan for these locations relative to the current working directory and home:
 | Location | Contents |
 |----------|----------|
 | `.work/node-cve/repos/` | Shallow repo clones |
-| `.work/node-cve/triage-YYYY-MM-DD/` | Triage reports and analysis |
+| `.work/node-cve/triage-YYYY-MM-DD/` | CVE triage reports and analysis |
+| `.work/node-bug/triage-YYYY-MM-DD/` | Bug triage reports |
+| `.work/node-rpm/` | Dist-git clones and Vagrant VM |
 | `~/.node-assistant/` | Cached team roster JSON files |
 
 For each location that exists, compute the size with `du -sh`.
 
 ### 2. Filter by age
 
-For `.work/node-cve/triage-*/` directories, parse the date from the directory
-name and skip directories newer than `--older-than` days. The `repos/` and
-`~/.node-assistant/` directories are always included (no age filter).
+For `.work/node-cve/triage-*/` and `.work/node-bug/triage-*/` directories,
+parse the date from the directory name and skip directories newer than
+`--older-than` days. The `repos/`, `.work/node-rpm/`, and `~/.node-assistant/`
+directories are always included (no age filter).
 
 ### 3. Show summary
 
@@ -54,9 +57,11 @@ Location                                          Size    Age
 $PWD/.work/node-cve/repos/                        1.2G    -
 $PWD/.work/node-cve/triage-2026-05-01/            45M     59 days
 $PWD/.work/node-cve/triage-2026-06-15/            52M     14 days (skipped, < 30 days)
+$PWD/.work/node-bug/triage-2026-05-01/            12M     59 days
+$PWD/.work/node-rpm/                              380M    -
 $HOME/.node-assistant/                            128K    -
 
-Total to remove: 1.25G
+Total to remove: 1.64G
 ```
 
 ### 4. Delete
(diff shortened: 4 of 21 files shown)


saschagrunert:
@harche @haircommander PTAL

plugins/node-bug/commands/triage.md
```diff
@@ -62,35 +65,14 @@ Designed for both interactive triage sessions and headless execution via `claude
...
-         "components", "labels",
-         "customfield_10689",
-         "customfield_10840",
-         "customfield_10847",
-         "customfield_10978",
-         "customfield_10020"
-       ]
-     }'
+   jira issue list -q "<constructed JQL>" \
+     --plain --no-headers --columns KEY,SUMMARY,COMPONENT,STATUS,ASSIGNEE,LABELS,PRIORITY
```
harche: COMPONENT isn't a valid column for `jira issue list`. The CLI's [ValidIssueColumns()](https://github.com/ankitpokhrel/jira-cli/blob/396933d554269f885d0f784c476b6cd09ae0c71e/internal/view/helper.go#L63-L78) only allows TYPE, KEY, SUMMARY, STATUS, ASSIGNEE, REPORTER, PRIORITY, RESOLUTION, CREATED, UPDATED and LABELS, and unknown columns are [silently skipped](https://github.com/ankitpokhrel/jira-cli/blob/396933d554269f885d0f784c476b6cd09ae0c71e/internal/view/issues.go#L176) rather than rejected. With `--no-headers` there is no header row to reveal the omission, so the output simply has no component data. Phase 2 step 1 routes every bug to its sub-team by component, so a default run has nothing to route on. The same latent issue exists in [node-cve's query skill](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-cve/skills/query-open-cves/SKILL.md#L31), though it doesn't depend on the value the way this command does.

plugins/node-bug/commands/triage.md
```diff
@@ -62,35 +65,14 @@ Designed for both interactive triage sessions and headless execution via `claude
...
-         "key", "summary", "status", "priority", "assignee",
-         "components", "labels",
-         "customfield_10689",
-         "customfield_10840",
-         "customfield_10847",
-         "customfield_10978",
-         "customfield_10020"
-       ]
-     }'
+   jira issue list -q "<constructed JQL>" \
```
harche: This needs `-p OCPBUGS` or a `project = OCPBUGS AND` prefix in the JQL. jira-cli seeds every list query with the configured default project ([NewJQL](https://github.com/ankitpokhrel/jira-cli/blob/396933d554269f885d0f784c476b6cd09ae0c71e/pkg/jql/jql.go#L27), called from [Issue.Get](https://github.com/ankitpokhrel/jira-cli/blob/396933d554269f885d0f784c476b6cd09ae0c71e/internal/query/issue.go#L58)), and [Raw()](https://github.com/ankitpokhrel/jira-cli/blob/396933d554269f885d0f784c476b6cd09ae0c71e/pkg/jql/jql.go#L180-L192) only drops that clause when the raw query itself contains a project filter. This JQL has none, so anyone whose CLI was initialized with default project OCPNODE silently runs `project="OCPNODE" AND filter = "Node Bugs" AND ...` and gets empty or undercounted results. node-cve avoids this because [its JQL starts with `project = OCPBUGS`](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-cve/skills/query-open-cves/SKILL.md#L31). Applies to the bucket queries and the workload query below too.

plugins/node-bug/commands/triage.md
```diff
@@ -107,26 +89,43 @@ Designed for both interactive triage sessions and headless execution via `claude
...
 
-   - **Release Blockers**: `"Release Blocker"` field value is "Approved"
-   - **Potential Blockers**: priority is "Blocker" AND `"Release Blocker"` is empty (Blocker priority set but no release blocker decision yet)
-   - **Customer Escalations**: `customfield_10978` (SFDC Cases Counter) is not null/empty, OR `customfield_10689` (Customer Impact) value is "Customer Escalated"
-   - **Component Regressions**: labels contain `component-regression`
-   - **Untriaged**: priority is "Undefined", OR `"Release Blocker"` is "Proposed", OR assignee is `aos-node@redhat.com`
-   - **Other**: all remaining bugs
+   JQL bucket conditions:
+   - **Release Blockers**: `AND "Release Blocker" = Approved`
+   - **Potential Blockers**: `AND priority = Blocker AND "Release Blocker" is EMPTY`
```
harche: These two bucket conditions disagree with the [Bug Triage Definitions table](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/skills/node/references/jira.md#L181-L182) that this step calls the source of truth. There, Blocker+ is `"Release Blocker" = Approved OR priority = Blocker` and Blocker? includes `"Release Blocker" = Proposed`. As written here, a Critical bug with Release Blocker set to Proposed lands in Untriaged instead of Potential Blockers, and a Blocker priority bug with an empty Release Blocker field never appears under Release Blockers. Worth either aligning these with the table or dropping the source of truth wording.

plugins/node-bug/.claude-plugin/plugin.json
```diff
@@ -1,14 +1,14 @@
 {
   "name": "node-bug",
   "description": "Node-specific bug triage, sub-team routing, and assignment suggestions for OpenShift Node team components",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
harche, line 4: This feels like more than a patch. The command drops the `JIRA_API_TOKEN` and curl auth path entirely and hard requires a configured jira CLI, so an existing headless setup that followed the old prerequisites now fails at the `jira me` gate (the [headless example](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-bug/commands/triage.md#L207) is kept), and the README also loses `--notify-slack`. [CONTRIBUTING.md](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/CONTRIBUTING.md#L39-L41) reserves MAJOR for breaking changes to existing commands, and this PR applies exactly that rule to node-team (0.18.1 to 1.0.0) for its jira CLI change. Suggest bumping this higher for consistency, or at least calling out the break.

plugins/node-team/commands/preflight.md
```diff
@@ -72,8 +72,8 @@ jira me
 ```
 
 - Pass: returns current user info
-- Fail: not installed or not configured (this is optional since the
-  `node-team` plugin uses `curl` directly, but `node-cve` requires it)
+- Fail: not installed or not configured (required by `node-cve` and
```
harche: Making this required cuts both ways. A user with only node-team installed now fails a required check for a CLI that node-team deliberately avoids ([jira.md](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/skills/node/references/jira.md#L3-L7) says to use curl directly). And in the other direction, checks 3 and 4 still hard require `JIRA_API_TOKEN` plus curl connectivity, which node-bug no longer uses anywhere after this PR, so a node-bug user with a working jira CLI fails preflight anyway. A note in the summary about which plugin needs which check would fix both.

plugins/node-team/commands/cleanup.md
```diff
@@ -54,27 +57,41 @@ Location                                          Size    Age
...
 If `--dry-run` is set, stop after the summary.
 
 Otherwise, ask the user for confirmation ("Remove these artifacts? [y/N]"),
-then delete with `rm -rf`. Report each deletion:
+then delete each location. For `.work/node-rpm/`, run `vagrant destroy -f`
+first to release libvirt VMs before removing the directory:
+
+```bash
+if command -v vagrant &>/dev/null && [ -f .work/node-rpm/Vagrantfile ]; then
+  cd .work/node-rpm && vagrant destroy -f; cd -
```
harche: If `vagrant destroy -f` fails here (libvirtd down, stale state), the `;` swallows the exit code and the `rm -rf` step right after still deletes the Vagrantfile and `.vagrant` metadata, orphaning exactly the libvirt domain the [note below](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/commands/cleanup.md#L96) says this avoids. A subshell keeps the cwd safe and makes failure visible: `(cd .work/node-rpm && vagrant destroy -f) && rm -rf .work/node-rpm/`.

harche (changes requested):
Thanks for this, the version cleanup and the family wide consistency sweep are welcome. I left a few comments inline, the two on the Phase 1 query are the important ones since both fail silently.

One bigger question first: is the jira CLI the right direction for this command at all? These runbooks are executed by the model, not typed by humans, so curl's verbosity costs us nothing. And curl is strictly more capable here: the old version fetched every bucket field and the SFDC counter in a single `search/jql` call, while the CLI cannot emit custom fields at all, which is what forces the four extra bucket queries and the per bug `--raw` fetches in this rewrite. Its column handling and default project behavior also caused the two silent failures flagged inline. Meanwhile [jira.md](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/skills/node/references/jira.md#L3-L7) in this same PR still declares curl the canonical approach for node-team, so the family stays split either way. Standardizing on curl and course correcting node-cve later seems like the more consistent end state, and it also drops a per machine `jira init` setup step for the whole team. Happy to hear the counterargument if there is one.

A few spots outside this diff that the consistency sweep missed:

* [overview.md](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/commands/overview.md#L47-L49) still says the roster set is core and dra only and that Kueue has no roster file, contradicting the updated [jira.md](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/skills/node/references/jira.md#L96) and node-bug's README. cleanup.md also points at overview as the roster resync path, so the kueue roster never gets cached that way.
* Compliance docs: the [dataflow.md local persistence table](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/docs/compliance/dataflow.md#L65-L69) doesn't list `.work/node-bug/` or `.work/node-rpm/` which cleanup.md now purges, and [user-guide.md](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-team/docs/compliance/user-guide.md#L9) still says the suite is just node-team and node-cve.
* node-bug's description in [marketplace.json](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/.claude-plugin/marketplace.json#L359) is missing the "for OpenShift Node team components" suffix that [plugin.json](https://github.com/openshift-eng/ai-helpers/blob/17d0ebdaa9afa5628a853a7f91b4a86f28d5d2f9/plugins/node-bug/.claude-plugin/plugin.json#L3) and docs/index.html have, and `sync_marketplace_versions.py` only syncs versions so it won't fix itself.

saschagrunert (commented):
Agreed on curl. The jira CLI's silent column dropping and default project injection are dealbreakers for a command that depends on component data for routing. Reverted node-bug to curl, which also eliminates the extra bucket queries and per-bug `--raw` fetches. node-cve stays on jira CLI for now since it doesn't depend on component values the same way.

Fixed the other consistency gaps you caught: overview.md now includes the Kueue roster file, dataflow.md and user-guide.md list node-bug and node-rpm, and the marketplace.json description matches plugin.json.

plugins/node-bug/commands/triage.md
saschagrunert: Reverted to curl POST with explicit field list. Components come back as structured JSON, no silent dropping.

plugins/node-bug/commands/triage.md
saschagrunert: Fixed by reverting to curl. The REST API executes the JQL as-is, no default project injection.

plugins/node-bug/commands/triage.md
saschagrunert: Aligned with the table. Release Blockers is now `"Release Blocker" = Approved OR priority = Blocker` (Blocker+) and Potential Blockers is `"Release Blocker" = Proposed OR (priority = Blocker AND "Release Blocker" is empty)` (Blocker?).

plugins/node-bug/.claude-plugin/plugin.json
saschagrunert: The breaking change (jira CLI requirement) is reverted, so the curl auth path is preserved. 0.0.2 covers the remaining fixes (bucket alignment, description wording). No breaking change to existing setups.

plugins/node-team/commands/preflight.md
saschagrunert: Updated. Jira CLI check now says "required by `node-cve`; `node-team` and `node-bug` use `curl` directly" and the summary table shows "PASS (node-cve only)".

plugins/node-team/commands/cleanup.md
saschagrunert: Fixed. Uses a subshell now and skips the `rm -rf` on failure:
```bash
(cd .work/node-rpm && vagrant destroy -f) || echo "WARNING: ..."
```

plugins/node-bug/README.md
```diff
@@ -29,21 +29,15 @@ Query open Node bugs, classify by severity and sub-team, suggest assignments, an
...
 - `--unassigned-only`: Show only untriaged or unassigned bugs
-- `--notify-slack`: Send summary to Slack (requires `SLACK_API_TOKEN` + `SLACK_CHANNEL` or `SLACK_WEBHOOK`)
 
 ## Prerequisites
 
-- `JIRA_API_TOKEN` environment variable (or macOS Keychain / Linux secret-tool)
-- `curl` (for Jira REST API and Slack)
-- Optional: `SLACK_API_TOKEN` + `SLACK_CHANNEL` or `SLACK_WEBHOOK` (for `--notify-slack`)
-- Optional: `~/.node-assistant/team-roster-{core,dra}.json` (for assignment suggestions)
+- `jira` CLI ([ankitpokhrel/jira-cli](https://github.com/ankitpokhrel/jira-cli)) configured with Jira credentials
```
harche: Leftover from the CLI version: the command is back on curl with the `JIRA_API_TOKEN` chain ([triage.md](https://github.com/openshift-eng/ai-helpers/blob/81943fa5efd1a2157e8fc94c8221498ad23266d5/plugins/node-bug/commands/triage.md#L31-L42)), so this prerequisite should return to `JIRA_API_TOKEN` (env var, macOS Keychain, or Linux secret-tool) plus `curl`. It is the last jira CLI reference left in the plugin.

plugins/node-team/docs/compliance/capabilities-inventory.md
```diff
@@ -6,7 +6,7 @@ Control ID: GLC-08
 
 | Tool | Purpose | Plugin |
 |------|---------|--------|
-| `jira` CLI | Query and comment on Jira issues | node-cve, node-team |
+| `jira` CLI | Query and comment on Jira issues | node-cve, node-bug, node-team |
```
harche: Now that node-bug is back on curl this row is stale in both directions: node-bug should come off the `jira` CLI row (per the updated [jira.md](https://github.com/openshift-eng/ai-helpers/blob/81943fa5efd1a2157e8fc94c8221498ad23266d5/plugins/node-team/skills/node/references/jira.md#L3-L8) only node-cve uses it), and the `curl` row below should gain node-bug since [triage.md](https://github.com/openshift-eng/ai-helpers/blob/81943fa5efd1a2157e8fc94c8221498ad23266d5/plugins/node-bug/commands/triage.md#L68) now calls the REST API directly.

plugins/node-team/docs/compliance/user-guide.md
```diff
@@ -11,6 +11,8 @@ component management. The suite consists of:
 - **node-team**: umbrella plugin for development setup, team overview, debugging,
   and shared reference data
 - **node-cve**: automated CVE triage with source code reachability analysis
+- **node-bug**: bug triage, sub-team routing, and assignment suggestions
+- **node-rpm**: RPM build and dist-git management
```
harche, line 16: The Quick Start right below still says "Install both plugins" and installs only node-team and node-cve, just under this list of four. Worth syncing the install snippet, and node-onboarding might belong in this list too if it is meant to cover the whole family.

plugins/node-bug/README.md
saschagrunert: Fixed. Switched to `JIRA_API_TOKEN` (env var, macOS Keychain, or Linux secret-tool) plus `curl`.

plugins/node-team/docs/compliance/capabilities-inventory.md
saschagrunert: Fixed. Removed node-bug and node-team from the `jira` CLI row, added both to the `curl` row.

plugins/node-team/docs/compliance/user-guide.md
saschagrunert: Fixed. Added node-onboarding to the plugin list and updated the install snippet to cover all five plugins.

PR #623, opened 2026-07-15
fix node-cve: add cross-team component validation safeguards to triage

asahay19:

Add cross-team component validation safeguards to `node-cve:triage` to prevent Node-specific CVE analysis from being posted to non-Node team Jira trackers.

On 2026-07-15, `/node-cve:triage --notify-jira` posted Node-specific reachability analysis for CVE-2026-27136 to non-Node OCPBUGS trackers (HyperShift, Storage, Networking/PTP, Installer, Monitoring, Cloud providers, Security, Windows Containers, etc.). Root cause: at posting time, trackers were looked up by CVE ID alone, without a component filter, so all of that CVE's  multi-team trackers were treated as a single batch to comment on.

This PR adds two independent layers of defense:

**`query-open-cves` :**
- Makes explicit that omitting `--component` means "all Node team components," never "no filter" — the `component in (...)` clause in the JQL query is now mandatory in every invocation, whether or not `--component` was passed.
- Adds a sanity check after querying: any returned tracker whose component is outside the Node team list is dropped and logged as a warning, to catch JQL construction bugs.

**`report-findings`:**
- Adds a mandatory "Node Team Component Safeguard" validation gate: immediately before posting any comment, each tracker's component is re-checked against the Node team list, independent of whatever Phase 1 already filtered. This is defense-in-depth so a stale cache, manual re-run, or ad-hoc script can't bypass it.
- Explicitly documents the anti-pattern that caused the incident (CVE-ID-only Jira search, ad-hoc "batch posting scripts" that bypass the normal filtered tracker list) so it isn't repeated.
- Adds a posting audit log (`posting-audit.log`) recording every tracker skipped and why, and surfaces a warning in the command summary output when any trackers are skipped.

Both safeguards reference the canonical Node team component list in the `node-team` shared reference (`plugins/node-team/skills/node/references/shared/components.md`) rather than duplicating a hardcoded list, so they can't drift out of sync as Node team components change.

Also bumps `node-cve` to `0.0.6` and syncs marketplace/docs metadata.

## Which issue(s) this PR fixes:

N/A — prompted by an internal incident (2026-07-15), not a filed issue.

- An earlier internal fix proposal suggested hardcoding a 3-component allowlist (`Node / CRI-O`, `Node / kubelet`, `Node / cri-tools`). That list is incomplete — the real Node team component list has ~14 Jira components (CPU manager, Memory manager, Topology manager, Device Manager, Kueue, etc.) plus `pscomponent:` label mappings (cadvisor, conmon, conmon-rs, cri-tools). This PR instead points both safeguards at the existing canonical shared reference so the safeguard itself doesn't become a new source of drift.
- These are Claude-instruction skills (Markdown), not executable code, so "validation" is enforced via explicit, repeated instructions plus a posting audit log for visibility — not a unit-testable function. Manual testing is recommended with a known multi-component CVE (e.g. CVE-2026-27136) to confirm only Node trackers receive comments and everything else is logged as skipped.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Strengthened CVE/Jira safeguards to ensure only validated Node team components are queried and notified.
  * Added clearer handling for mismatched trackers, including user-visible warnings and skip behavior.
  * When Jira notifications are enabled, now generates a `posting-audit.log` with posted vs. skipped tracker details.

* **Documentation**
  * Updated `node-cve:triage` and related skills with explicit, non-optional component scoping and re-validation guidance.

* **Chores**
  * Updated the `node-cve` plugin version to `0.0.6`.

### DIFF
--- a/plugins/node-cve/commands/triage.md
+++ b/plugins/node-cve/commands/triage.md
@@ -29,7 +29,7 @@ Designed for both interactive use and headless execution via `claude --print`.
 ### Phase 0: Setup and Argument Parsing
 
 1. **Parse Arguments**
-   - `--component <name>`: Filter to a specific OCPBUGS component (e.g., "Node / CRI-O"). Optional.
+   - `--component <name>`: Filter to a single specific OCPBUGS component (e.g., "Node / CRI-O"). Optional. **If omitted, the command includes ALL Node team components — and only Node team components. It never queries or posts to components outside the Node team.** See the [node-team shared components reference](../../node-team/skills/node/references/shared/components.md) for the canonical list.
    - `--notify-jira`: Post analysis results as comments on Jira tracker issues. Default: off.
    - `--notify-slack`: Send summary to Slack. Requires either `$SLACK_API_TOKEN` + `$SLACK_CHANNEL` (enables threading) or `$SLACK_WEBHOOK` (simpler, no threading). Default: off.
    - `--days N`: Only include CVEs created or updated in the last N days. Default: all open.
@@ -55,6 +55,8 @@ Designed for both interactive use and headless execution via `claude --print`.
 - **Input**: Optional `--component` filter, optional `--days` filter
 - **Output**: Deduplicated list of CVEs with metadata
 
+**CRITICAL SAFEGUARD:** The `component in (...)` filter below is mandatory and must never be omitted, regardless of whether `--component` was passed. This is the first of two layers of defense against cross-team contamination — the second is the posting-time re-validation in [report-findings](../skills/report-findings/SKILL.md) Phase 3, Step 2. Both layers exist because a prior incident (2026-07-15) showed that CVE analysis can otherwise be posted to 200+ trackers belonging to other OpenShift teams.
+
 **Steps:**
 
 1. Build the JQL query using the CVE-tracked component list from the [node-team shared components reference](../../node-team/skills/node/references/shared/components.md) (the full Jira component list plus Driver Toolkit and Machine Config Operator):
@@ -63,7 +65,7 @@ Designed for both interactive use and headless execution via `claude --print`.
    jira issue list -q "project = OCPBUGS AND type = Vulnerability AND component in (<components from shared reference>) AND status not in (Closed, Done, Verified)" --plain --no-headers --columns KEY,SUMMARY,COMPONENT,STATUS,ASSIGNEE,LABELS
    ```
 
-   If `--component` is specified, replace the component list with the single component.
+   If `--component` is specified, replace the component list with the single component (still using `component in (...)`, never an unfiltered query).
    If `--days N` is specified, add `AND updated >= -${N}d` to the query.
 
 2. Parse results and extract CVE IDs from summaries (regex: `CVE-[0-9]{4}-[0-9]+`).
@@ -218,7 +220,9 @@ The per-branch results are preserved in the report so reviewers can see which ve
    **Recommended action:** <update dependency / apply patch / monitor / investigate>
    ```
 
-2. **If `--notify-jira`**: For each CVE, post a comment on ALL its tracker issues. Each tracker receives the analysis result specific to its OCP version/branch.
+2. **If `--notify-jira`**: For each CVE, post a comment on ALL its **Node-team-validated** tracker issues. Each tracker receives the analysis result specific to its OCP version/branch.
+
+   **Before posting anything, re-validate every tracker's component against the Node team component list — do not rely solely on Phase 1's filtering.** See [report-findings](../skills/report-findings/SKILL.md) "Node Team Component Safeguard" and Step 2 for the mandatory validation logic and audit logging. Skip (never post to) any tracker whose component is not a Node team component, and record it in the posting audit log.
 
    Use Atlassian wiki markup (not Markdown), matching the format in [report-findings](../skills/report-findings/SKILL.md) Step 2. The comment includes per-branch results across all analyzed versions so reviewers can see the full picture.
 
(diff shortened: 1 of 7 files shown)


plugins/node-cve/skills/report-findings/SKILL.md
```diff
@@ -72,7 +89,28 @@ List CVEs that are Reachable or Uncertain with unassigned owners. These need imm
...
 
-For each unique CVE, post a comment on ALL its tracker issues. Each tracker issue receives the analysis result for its specific OCP version/branch (not a blanket result). Use Atlassian wiki markup (not Markdown):
+**VALIDATION (MANDATORY, before posting anything):** For each unique CVE, take its `tracker_keys` list from Phase 1 (`query-open-cves`) and re-validate every tracker's component immediately before posting — do not trust cached or upstream filtering alone:
+
+```bash
+# component_ok() checks a tracker's component/labels against the canonical
+# Node team component list from the shared components reference (link above),
+# NOT a hardcoded list.
+for tracker_key in $TRACKER_KEYS; do
+  component=$(jira issue view "$tracker_key" --plain --no-headers --columns COMPONENT | tail -1)
```
saschagrunert, line 99: The two `jira issue view` calls (lines 99-100) each make a separate API call per tracker. These can be combined into one:

```bash
output=$(jira issue view "$tracker_key" --plain --no-headers --columns COMPONENT,LABELS | tail -1)
```

This halves the validation API calls. Also, the loop has no `sleep 1` between calls, but the existing rate limit instruction (line 159) says "sleep 1 second between Jira API calls." Since the agent will likely follow this pseudocode literally, adding `sleep 1` inside the loop would prevent 429s. For a CVE with 30 trackers, the current pseudocode fires 60 unthrottled API calls before posting even starts.

plugins/node-cve/skills/report-findings/SKILL.md
```diff
@@ -72,7 +89,28 @@ List CVEs that are Reachable or Uncertain with unassigned owners. These need imm
...
+  if ! component_is_node_team "$component" "$labels"; then
+    echo "⚠️  SKIPPING $tracker_key: component '$component' is not a Node team component" | tee -a "$SKIPPED_LOG"
+    continue
+  fi
+
+  # Component validated — proceed with posting for $tracker_key
+done
+```
+
+A component is considered a Node team component only if it matches an entry in the "Jira Components (OCPBUGS)" list (plus Driver Toolkit, Machine Config Operator) or one of the `pscomponent:` labels in the shared components reference. If a tracker's component cannot be confidently classified as Node team, **skip it and log the reason** — never post "just in case." This check must run even when `--component` was passed, and even if Phase 1 already filtered by component, since this is the last line of defense before an irreversible write to another team's tracker.
```
saschagrunert: The `pscomponent:` label check creates a false-pass risk in the posting-time validation. A tracker with component "Security" (non-Node) but label `pscomponent:cri-o` would pass this check and get posted to, which is exactly the cross-team contamination this PR prevents.

Phase 1 already filters by `component in (...)`, so any tracker reaching Phase 3 should have a Node team component. The defense-in-depth re-validation should verify the COMPONENT field only, not offer pscomponent labels as an alternative pass condition. The pscomponent labels are for Phase 2 repo mapping (which repo to clone), not for determining tracker ownership.

plugins/node-cve/skills/report-findings/SKILL.md
```diff
@@ -311,16 +365,19 @@ Ensure all generated files exist under `.work/node-cve/triage-YYYY-MM-DD/`:
   "report_path": ".work/node-cve/triage-2026-05-20/report.md",
   "jira_comments_posted": 45,
   "jira_comments_failed": 0,
+  "jira_trackers_skipped_non_node_component": 0,
   "slack_notified": true,
   "artifacts": [
     ".work/node-cve/triage-2026-05-20/report.md",
-    ".work/node-cve/triage-2026-05-20/cves.json"
+    ".work/node-cve/triage-2026-05-20/cves.json",
+    ".work/node-cve/triage-2026-05-20/posting-audit.log"
```
saschagrunert: `posting-audit.log` is listed unconditionally in the artifacts array, but it's only written when `--notify-jira` is used (see Step 2, line 163). A run without `--notify-jira` would have the agent report a nonexistent file. Either note this artifact is conditional on `--notify-jira`, or move it out of the example and into a note below it.

plugins/node-cve/commands/triage.md
```diff
@@ -29,7 +29,7 @@ Designed for both interactive use and headless execution via `claude --print`.
 ### Phase 0: Setup and Argument Parsing
 
 1. **Parse Arguments**
-   - `--component <name>`: Filter to a specific OCPBUGS component (e.g., "Node / CRI-O"). Optional.
+   - `--component <name>`: Filter to a single specific OCPBUGS component (e.g., "Node / CRI-O"). Optional. **If omitted, the command includes ALL Node team components — and only Node team components. It never queries or posts to components outside the Node team.** See the [node-team shared components reference](../../node-team/skills/node/references/shared/components.md) for the canonical list.
```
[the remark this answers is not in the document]
asahay19: I think, this finding is a false positive .The overlap-detection script compared PR #623 against itself (#623 is this PR), so there's nothing external to consolidate with. There's no other open PR touching these files. Marking resolved.

plugins/node-cve/skills/report-findings/SKILL.md
asahay19: Thanks sascha. I have Added sleep 1 between validation calls and made posting-audit.log conditional on --notify-jira in the artifacts list.

plugins/node-cve/skills/report-findings/SKILL.md
asahay19: Addressed in the latest commit, please check.

saschagrunert (approved):

saschagrunert:
Images CI seems to be broken all over the place, unfortunately I can't fix that now.

PR #489, opened 2026-05-20
HPNEX-16: feat: add bundles with openshift-developer meta-package

stbenjam:
## What this PR does / why we need it:

Adds a bundles/ directory with curated sets of APM packages. Includes the openshift-developer bundle and documents marketplace registration for plugins that depend on anthropics/claude-plugins-official.

Requires https://github.com/microsoft/apm/pull/1422 first

## Which issue(s) this PR fixes:

N/A

## Checklist:
- [X] Subject and description added to both, commit and PR.
- [X] Relevant issues have been referenced.
- [X] This change includes docs.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guide for APM bundles covering available options, registration prerequisites, and installation instructions for both global and project-scoped deployments, plus guidance for creating custom bundles.

* **New Features**
  * Introduced the openshift-developer bundle with pre-configured dependencies and integration settings.

PR #626, opened 2026-07-16
Add 3CMO false positive eval for payload-analysis

theobarberbany:
## Summary

Adds 3 eval cases for payload-analysis targeting the CCCMO #488 false positive, plus judge improvements to catch hallucinated evidence and temporal-correlation-only attribution.

### Eval cases

- **case-015** (`5.0.0-0.nightly-2026-07-13-041249`) — Rejected nightly with full JUnit data. MCO root cause is obvious from test results. Tests correct attribution when evidence is available.
- **case-016** (`5.0.0-0.ci-2026-07-11-223154`) — Original CI payload as the agent saw it: empty JUnit, no build logs. Tests whether the agent fabricates evidence when data is missing. This is the exact data condition that produced the false positive.
- **case-017** (`5.0.0-0.ci-2026-07-11-223154-backfilled`) — Same CI payload with JUnit backfilled from GCS. Tests correct attribution with the data that should have been there originally.

### Judge improvements

- `analysis_quality`: added hard rules — cap at 2 if report concludes a `must_not_conclude` verdict or cites fabricated error strings. Score definitions now reference `distractors` and `discriminating_signal`.
- `revert_scoring_accuracy`: added hard rules for false-positive cases — cap at 1 if `expected_candidates` is empty and a `must_not_conclude` PR is recommended for revert. Score definitions now cover no-candidate cases.
- Dataset schema updated to document `key_evidence`, `discriminating_signal`, `distractors`, `must_not_conclude` (borrowed from prow-job-analysis eval).

### Snapshot script fix

- `payload_snapshot.py`: failed blocking jobs missing JUnit or build logs now get a `data_gaps` field in `summary.json` (e.g. `data_gaps: ["junit", "build_log"]`) instead of silently omitting the data.

### Snapshot data

Depends on https://github.com/stbenjam/historical-payload-data/pull/2 for the 3 payload snapshots.

### Context

The payload agent attributed GCP micro upgrade failures to CCCMO PR #488 at 100% confidence ([revert PR #489](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/489), closed without merging). It fabricated the error string "Deployment cloud-controller-manager rollout not complete" — a klog.V(2) debug line from the PR diff that was never emitted. The actual root cause was MCO: "Operator upgrade machine-config" failed 10/10, while "verify operator conditions cloud-controller-manager" passed 2/2.


stbenjam:
Heads-up from the payload-analysis A/B work in #659/#660: `agent-eval-harness` exposes the case directory name to the model, so descriptive names leak the expected diagnosis. We have renamed the existing dataset to opaque `case-NNN` directories and put the human mapping in `plugins/ci/evals/cases/payload-analysis/README.md`. Before this PR merges, could cases 015–017 likewise be named only `case-015`, `case-016`, and `case-017`? The index in #659/#660 already reserves those IDs for these scenarios.

PR #630, opened 2026-07-17
payload-analysis: do not recommend revert or force-reject for kube rebase version skew

sosiouxme:
## Summary

When a Kubernetes rebase merges in `openshift/kubernetes`, there is an expected lag of hours while the kubelet is rebuilt against the new source and delivered via an updated RHCOS image. During this window, kubelet version skew (kubelet on an older kube version than kube-apiserver) causes widespread test failures across all platforms.

The payload agent currently treats this like any other PR-caused regression: it scores the rebase PR at high confidence and recommends reverting it. This is incorrect — the kubelet binary is built *from* `openshift/kubernetes`, so the rebase must remain for the kubelet to pick up the new version. Reverting would prevent the kubelet from ever being updated.

## Changes

Two additive sections in `plugins/ci/skills/payload-analysis/SKILL.md`:

**Step 6.2 — Special case: Kubernetes rebase version skew**
- When the candidate PR is a kube rebase AND failures show kubelet version skew, do NOT propose a revert even at score >= 85
- Classify as "transient build lag" and recommend monitoring for the rebuilt kubelet to land via RHCOS

**Step 6.4 — Guard: kubelet version skew from a Kubernetes rebase**
- Force-accepting is inappropriate (real test failures, not infra flakes)
- Force-rejecting is also inappropriate (just accelerates the next payload into the same skew)
- Correct action: wait for the RHCOS with the rebuilt kubelet

## Context

This was identified during analysis of `5.0.0-0.nightly-2026-07-17-034502` where the agent incorrectly recommended reverting `openshift/kubernetes#2653` (Kubernetes 1.36.2 rebase). Feedback from TRT in https://redhat-internal.slack.com/archives/C02K89U2EV8/p1752756363338069.

---
same initial content as https://github.com/openshift-eng/ai-helpers/pull/627 but that one is hung up getting the eval case committed

## Summary by CodeRabbit

* **Documentation**
  * Updated payload analysis guidance for Kubernetes rebases affected by kubelet version skew.
  * These cases are now classified as test failures with a transient build lag root cause.
  * Added guidance to wait for the updated RHCOS image rather than accept or reject the change immediately.

* **Chores**
  * Updated the CI plugin version to 0.0.70 across marketplace metadata and documentation.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -88,7 +88,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-      "version": "0.0.69",
+      "version": "0.0.70",
       "category": "ci",
       "keywords": [
         "prow",
--- a/docs/index.html
+++ b/docs/index.html
@@ -518,7 +518,7 @@ footer a { color: var(--primary); }
     {
       "name": "ci",
       "description": "Tools for working with OpenShift CI and analyzing Prow job results",
-      "version": "0.0.69",
+      "version": "0.0.70",
       "has_readme": true,
       "commands": [
         {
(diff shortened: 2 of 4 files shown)


smg247 (commented):
/lgtm

PR #633, opened 2026-07-21
fix(ci): bash syntax error in setup-prow-job-analysis.sh

stbenjam:
## Summary

- Fixes bash syntax error (`syntax error near unexpected token '('`) in `setup-prow-job-analysis.sh:62`
- The heredoc inside `$(...)` command substitution caused bash to misparse Python parentheses as shell subshell syntax
- Moves the heredoc out of `$(...)` and captures the exit code separately

Unblocks rehearsal of `eval-prow-job-analysis` in openshift/release#81386.

## Test plan

- [x] `bash -n` passes on the fixed script

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics when probing access to the public test-results storage.
  * Clearly reports whether the storage is reachable, including the captured error output when it is not.
  * Continues CI setup with the existing best-effort pre-warming behavior even if the storage check fails.
* **Chores**
  * Updated CI plugin version metadata to the latest patch.

### DIFF
--- a/plugins/ci/.claude-plugin/plugin.json
+++ b/plugins/ci/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "ci",
   "description": "Tools for working with OpenShift CI and analyzing Prow job results",
-  "version": "0.0.74",
+  "version": "0.0.75",
   "author": {
     "name": "github.com/openshift-eng"
   }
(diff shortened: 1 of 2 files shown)


cblecker (approved):
Bash fix looks correct — heredoc inside `$(...)` is a real parsing issue, and the replacement pattern properly handles `set -euo pipefail` while preserving error reporting. LGTM.

PR #650, opened 2026-07-27
payload-snapshot: log podman info output when nested podman check fails

petr-muller:
The RHCOS RPMDB collector runs 'podman info' as a preflight check for nested podman support, but only reported a bare pass/fail. When the check fails, the underlying runtime/storage/namespace error from podman is now logged (stdout and stderr, or the OSError/timeout detail) so failures under nested podman can actually be diagnosed instead of just "podman is not available".

## Summary by CodeRabbit

* **New Features**
  * Updated the CI plugin to version 0.0.73 across the marketplace and documentation.

* **Bug Fixes**
  * Improved Podman environment checks with clearer diagnostic details when commands fail, time out, or cannot be started.
  * Podman availability checks continue to report unsuccessful results reliably without interrupting processing.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -88,7 +88,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-      "version": "0.0.72",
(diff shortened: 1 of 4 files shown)


smg247 (commented):
/lgtm

PR #654, opened 2026-07-28
TRT-2612: Fix nested podman subuid/subgid range for hostUsers: false pods

petr-muller:
OpenShift CI pods running with hostUsers: false only get 65536 UIDs/GIDs (0-65535) mapped into the pod's user namespace. The entrypoint's hardcoded subuid/subgid range (100000:65536) falls entirely outside that window, so podman's newuidmap/newgidmap invocations fail with EPERM (`newuidmap: write to uid_map failed: Operation not permitted`), breaking nested rootless podman.

This computes the subuid/subgid start and count dynamically from the runtime UID (e.g. uid 1000 -> 1001:64535), so the range always fits within 0-65535. Env var overrides are preserved.

Trade-off: nested containers get ~64.5k subordinate IDs instead of 65536, so images with files owned by UIDs above ~64535 won't run nested. That's inherent to the 65536-wide pod user namespace.

Exercised by re-running the rehearsal on openshift/release PR #81145 after merge.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rootless container startup in OpenShift environments by dynamically configuring user and group ID mappings.
  * Prevented mapping ranges from exceeding pod user-namespace limits, improving compatibility with restricted containers.

### DIFF
--- a/images/nested-podman-entrypoint.sh
+++ b/images/nested-podman-entrypoint.sh
@@ -46,8 +46,21 @@ fi
 
 # Create subuid/gid entries for rootless podman
 USER_NAME="$(whoami)"
-SUBID_START="${SUBID_START:-100000}"
-SUBID_COUNT="${SUBID_COUNT:-65536}"
+# In OpenShift CI, the pod runs with hostUsers: false, which maps only 65536
+# UIDs/GIDs (0-65535) into the pod's user namespace. The host rootless-podman
+# default of 100000:65536 falls outside that range, so newuidmap/newgidmap
+# fail with EPERM. Compute the range dynamically so it always fits in 0-65535.
+# Outside hostUsers: false (e.g. a normal OpenShift pod with a large
+# SCC-assigned arbitrary UID), that computation goes negative or zero, so
+# fall back to the host rootless-podman default in that case.
+SUBID_START_DEFAULT=$(( $(id -u) + 1 ))
+SUBID_COUNT_DEFAULT=$(( 65536 - SUBID_START_DEFAULT ))
+if (( SUBID_COUNT_DEFAULT <= 0 )); then
+  SUBID_START_DEFAULT=100000
+  SUBID_COUNT_DEFAULT=65536
+fi
+SUBID_START="${SUBID_START:-$SUBID_START_DEFAULT}"
+SUBID_COUNT="${SUBID_COUNT:-$SUBID_COUNT_DEFAULT}"
 echo "${USER_NAME}:${SUBID_START}:${SUBID_COUNT}" > /etc/subuid
 echo "${USER_NAME}:${SUBID_START}:${SUBID_COUNT}" > /etc/subgid
 


petr-muller:
/retitle [TRT-2612](https://redhat.atlassian.net/browse/TRT-2612): Fix nested podman subuid/subgid range for hostUsers: false pods

images/nested-podman-entrypoint.sh
```diff
@@ -46,8 +46,12 @@ fi
...
 # Create subuid/gid entries for rootless podman
 USER_NAME="$(whoami)"
-SUBID_START="${SUBID_START:-100000}"
-SUBID_COUNT="${SUBID_COUNT:-65536}"
+# In OpenShift CI, the pod runs with hostUsers: false, which maps only 65536
+# UIDs/GIDs (0-65535) into the pod's user namespace. The host rootless-podman
+# default of 100000:65536 falls outside that range, so newuidmap/newgidmap
+# fail with EPERM. Compute the range dynamically so it always fits in 0-65535.
+SUBID_START="${SUBID_START:-$(( $(id -u) + 1 ))}"
+SUBID_COUNT="${SUBID_COUNT:-$(( 65536 - SUBID_START ))}"
```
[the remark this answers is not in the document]
petr-muller: Skipping this: the UID=65535 (and above) scenario can't occur in practice here. OpenShift's `hostUsers: false` pod user namespace only maps UIDs 0-65535, so the runtime UID podman actually runs as is always well below that boundary (e.g. 1000 in this image). Adding numeric/bounds validation for the override path is reasonable in the abstract, but the script doesn't validate any of its other overridable env vars either, so I don't think this one warrants special-casing. Leaving as-is.

petr-muller-reviewer: **[blocking]** The 0-65535 ceiling is a property of `hostUsers: false` pods specifically, not of OpenShift pods in general. `Dockerfile.nested-podman` (which uses this same entrypoint) notes elsewhere that OpenShift CI also runs with a random, typically large arbitrary UID (no pod-level user-namespace ceiling in that case). For such a UID, `SUBID_START = uid+1` is huge and `SUBID_COUNT = 65536 - SUBID_START` goes deeply negative, producing a malformed `/etc/subuid`/`/etc/subgid` entry (e.g. `claude:1000680001:-1000614465`) where the old fixed `100000:65536` worked fine. This risks regressing nested podman for the normal (non-`hostUsers: false`) case while fixing the `hostUsers: false` case.

images/nested-podman-entrypoint.sh
```diff
@@ -46,8 +46,12 @@ fi
 
 # Create subuid/gid entries for rootless podman
 USER_NAME="$(whoami)"
-SUBID_START="${SUBID_START:-100000}"
-SUBID_COUNT="${SUBID_COUNT:-65536}"
+# In OpenShift CI, the pod runs with hostUsers: false, which maps only 65536
+# UIDs/GIDs (0-65535) into the pod's user namespace. The host rootless-podman
+# default of 100000:65536 falls outside that range, so newuidmap/newgidmap
+# fail with EPERM. Compute the range dynamically so it always fits in 0-65535.
+SUBID_START="${SUBID_START:-$(( $(id -u) + 1 ))}"
```
petr-muller-reviewer: **[nit]** If the runtime UID is exactly 65535, `SUBID_COUNT` computes to 0, producing an empty/invalid subuid range. Likely rare but undocumented; could clamp to a minimum or note the assumption.

petr-muller-reviewer (changes requested):
**Verdict: request changes.** The `hostUsers: false` fix itself is correct, but the new formula is applied unconditionally and appears to assume the runtime UID is always small. The same entrypoint/image is also used in contexts with large arbitrary OpenShift-assigned UIDs (per `Dockerfile.nested-podman`'s own comment: "In CI, OpenShift uses a random UID, but in the root group"), where this change risks producing a negative subuid/subgid count and regressing a previously-working case.

**Checked and fine:**
- Arithmetic ordering: `SUBID_COUNT` is computed after `SUBID_START` is resolved, so it correctly reflects the actual start value used.
- Env var overrides preserved for both `SUBID_START` and `SUBID_COUNT`.
- Example math in the PR description (`uid 1000 -> 1001:64535`) is correct.
- No pre-existing test coverage for this script; this PR doesn't remove any.

**Open questions:**
- Is this entrypoint/image ever run in a pod without `hostUsers: false` (i.e. with a large arbitrary OpenShift-assigned UID)? If yes, does the formula need to detect/branch on that?
- Was the previously-working (non-`hostUsers: false`) nested-podman path re-tested after this change, or only the new `hostUsers: false` rehearsal?

images/nested-podman-entrypoint.sh
petr-muller: Good catch — addressed. Added a fallback: the dynamic hostUsers:false-scoped formula is only used when it computes a positive SUBID_COUNT; otherwise it falls back to the old fixed 100000:65536 range. This covers the general OpenShift arbitrary-UID case (large SCC-assigned UID, no pod-level userns ceiling) without needing to explicitly detect which pod mode is active.

Confirmed this image isn't deployed anywhere today, so I can't point to prior working behavior for that range — but the entrypoint can't fail/crash from this section regardless of UID or subuid content (the echo writes always succeed, and /etc/subuid is only ever consulted by subordinate-ID-aware tools like newuidmap/podman when something actually tries to build a rootless user namespace). So outside hostUsers:false, this at minimum won't break anything; nested podman there remains unverified either way, same as before this PR.

images/nested-podman-entrypoint.sh
petr-muller: Resolved as a side effect of the fallback added for the sibling comment above — UID=65535 now also hits the SUBID_COUNT_DEFAULT <= 0 guard and falls back to 100000:65536 instead of producing a zero-count range.

petr-muller-reviewer (approved):
Commit `366b528c` added a `SUBID_COUNT_DEFAULT <= 0` fallback to the old fixed `100000:65536` range whenever the dynamic uid-based computation would go zero or negative. This resolves the blocking concern (large arbitrary SCC-assigned UID in non-`hostUsers: false` pods no longer produces a negative subuid/subgid count) and the boundary nit (UID 65535 now also falls back cleanly). No open findings remain.

PR #658, opened 2026-07-30
fix(eval): use setup script for plugin deps, widen dependency ranges

enxebre:
## Summary

- Removes plugin install from `run-solve.sh` — it didn't work because each `claude -p` subprocess is a fresh session
- Adds `setup-solve-eval.sh` that installs plugins via marketplace before the eval runs (called via `EVAL_SETUP_SCRIPT` in the release repo CI config)
- Changes dependency ranges from `^` (caret) to `>=` in `plugin.json` — caret ranges on `0.x.y` versions only allow patch bumps, which breaks when dependent plugins are bumped to newer minor versions

## Why

The jira-solve eval uses `runner.type: cli` which spawns separate `claude -p` processes. Plugin installs inside `run-solve.sh` don't propagate to those inner sessions. The setup script runs before the eval starts, installing plugins system-wide so all subprocesses can find them.

The `^0.0.52` constraint rejects `0.0.77` because semver caret on `0.x` is patch-only (`>=0.0.52 <0.0.53`). Using `>=` allows any newer version.

## Companion PR

openshift/release#82470 — adds `EVAL_SETUP_SCRIPT` pointing to this setup script

## Test plan

- [x] `make lint` passes (A+)
- [ ] Release repo rehearsal with setup script

## Summary by CodeRabbit

* **New Features**
  * Added automated setup and validation for the OpenShift Developer plugin in evaluation environments.
* **Improvements**
  * Updated the plugin bundle to version 1.1.14.
  * Relaxed dependency requirements to support compatible newer versions.
  * Simplified evaluation startup and execution by removing manual plugin directory configuration.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -88,7 +88,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-"version": "0.0.78",
+      "version": "0.0.78",
       "category": "ci",
       "keywords": [
         "prow",
@@ -484,7 +484,7 @@
       "name": "openshift-developer",
       "source": "./plugins/openshift-developer",
       "description": "Bundle of curated plugins, skills, and MCP servers useful to any OpenShift engineer",
-      "version": "1.1.13",
+      "version": "1.1.14",
       "category": "bundle",
       "keywords": [
         "openshift",
--- a/plugins/openshift-developer/.claude-plugin/plugin.json
+++ b/plugins/openshift-developer/.claude-plugin/plugin.json
@@ -1,16 +1,16 @@
 {
   "name": "openshift-developer",
   "description": "Bundle of curated plugins, skills, and MCP servers useful to any OpenShift engineer",
-  "version": "1.1.13",
+  "version": "1.1.14",
   "author": {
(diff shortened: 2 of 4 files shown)


docs/index.html
```diff
@@ -608,14 +608,6 @@ <h1>ai-helpers</h1>
...
         },
-        {
-          "name": "create-symptom",
-          "full_name": "ci:create-symptom",
-          "description": "Interactively create (or update/delete) a Sippy Symptom that auto-labels a known CI failure pattern",
-          "description_html": "Interactively create (or update/delete) a Sippy Symptom that auto-labels a known CI failure pattern",
-          "synopsis": "/ci:create-symptom [description-of-failure]",
-          "body_html": "The \u003ccode\u003eci:create-symptom\u003c/code\u003e command creates a Sippy Symptom — a known-failure signature made of an artifact file pattern and a matcher that, when it matches a CI job run, automatically applies human-readable Labels (like \u003ccode\u003eInfraFailure\u003c/code\u003e) so future failures are recognized without re-debugging. The command walks through an interactive workflow: check for duplicate symptoms, verify (or create) the target labels, confirm the exact payload with the user, then create. It also handles updating or deleting an existing symptom when the user asks."
-        },
```
smg247, line 618: why are these commands being removed?

PR #668, opened 2026-08-03
payload-snapshot: extract RPM changelogs for packages changed vs. baseline

petr-muller:
Reconstructs the target-vs-baseline RPM version diff by aggregating the per-hop rhcos_changes already fetched for every payload in the chain (no dedicated API call), then extracts each changed package's `rpm --changelog` from the already-downloaded target RPMDB, trimmed to entries newer than the old version on a best-effort basis. Results are referenced from summary.json as rpm_changelogs[], and the whole step is gated behind a new --no-rpm-changelogs flag plus an rpm CLI preflight check.

## Summary by CodeRabbit

* **New Features**
  * Payload snapshots can optionally include RPM package changelogs for packages changed between baseline and target versions.
  * Snapshot summaries include changelog file paths and metadata, with automatic handling when RPM data is unavailable.
  * Added an option to disable RPM changelog collection.

* **Documentation**
  * Documented changelog extraction, trimming, generated files, summary fields, and skip conditions.

* **Chores**
  * Updated the CI plugin version to 0.0.80.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -41,7 +41,7 @@
         "repo": "rh-uxd/ai-helpers",
         "path": "plugins/patternfly"
       },
-      "description": "PatternFly development tools from the UXD team — component scaffolding, design guidance, auditing, migration, test generation, and an MCP server.",
+      "description": "PatternFly development tools from the UXD team \u2014 component scaffolding, design guidance, auditing, migration, test generation, and an MCP server.",
       "category": "development",
       "keywords": [
         "patternfly",
@@ -105,7 +105,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-      "version": "0.0.79",
+      "version": "0.0.80",
       "category": "ci",
       "keywords": [
         "prow",
--- a/plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
+++ b/plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
@@ -1578,13 +1578,15 @@ class SummaryGenerator:
 
     def __init__(self, base_dir: str, chain: list[str], target_tag: str,
                  tag: "PayloadTag", streaks: Optional[dict] = None,
-                 rpmdb_data: Optional[dict[str, list[dict]]] = None):
+                 rpmdb_data: Optional[dict[str, list[dict]]] = None,
+                 rpm_changelogs: Optional[list[dict]] = None):
         self.base_dir = base_dir
         self.chain = chain
         self.target_tag = target_tag
         self.tag = tag
         self.streaks = streaks or {}
         self.rpmdb_data = rpmdb_data or {}
+        self.rpm_changelogs = rpm_changelogs or []
 
     def generate(self) -> None:
         target_dir = os.path.join(self.base_dir, self.target_tag)
@@ -1664,6 +1666,9 @@ class SummaryGenerator:
                 {**entry, "rpmdb": f"{self.target_tag}/rpmdb/{entry['tag']}/rpmdb.sqlite"}
                 for entry in target_rpms
             ]
+
+        if self.rpm_changelogs:
+            summary["rpm_changelogs"] = self.rpm_changelogs
 
         # Surface any data that could not be collected.  An empty snapshot
         # section must never be silently indistinguishable from a clean one.
@@ -1782,6 +1787,8 @@ class SummaryGenerator:
             "    rpmdb/                    # RPMDB from RHCOS images",
             "      rhel-coreos/           # queryable with rpm --dbpath",
             "        rpmdb.sqlite",
+            "        changelogs/          # RPM changelogs vs. baseline",
+            "          <package>.txt",
             "      rhel-coreos-10/",
             "        rpmdb.sqlite",
             "  <older-payload-tag>/      # Each prior payload in the chain",
@@ -1828,6 +1835,13 @@ class SummaryGenerator:
             "- `rhcos_rpms[]` — RPMDB per RHCOS variant for the target",
             "  payload: `tag`, `name`, `pullspec`, `rpmdb` (relative path",
             "  to rpmdb.sqlite — queryable with rpm --dbpath <dir> or sqlite3)",
+            "- `rpm_changelogs[]` — RPM changelogs for packages changed",
+            "  vs. the baseline: `variant`, `package`, `old`, `new`,",
+            "  `changelog` (relative path to the extracted, trimmed",
+            "  `rpm --changelog` output), `boundary_found`, `trim_method`",
+            "  (`rpmdb_diff` — exact, diffed against the baseline's own",
+            "  rpmdb; `version_match` — heuristic version-string cut;",
+            "  `full_dump` — untrimmed, neither approach found a boundary)",
             "",
         ])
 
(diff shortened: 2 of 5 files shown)


petr-muller:
Needs https://github.com/openshift/release/pull/81145

plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
```diff
@@ -2676,6 +2799,28 @@ def _extract_rhcos_changes(changelog: dict) -> list[dict]:
...
+    conventionally end each entry's header line with "- <version>". Cut
+    there if found; otherwise return the full changelog unchanged, since a
+    wrong guess is worse than showing extra history.
+    """
+    bare_old = old_version.split(":", 1)[-1]  # strip epoch prefix, if any
+    entries = re.split(r"(?=^\* )", changelog_text, flags=re.MULTILINE)
+    kept = []
+    for entry in entries:
+        first_line = entry.splitlines()[0] if entry.strip() else ""
+        if bare_old and bare_old in first_line:
```
petr-muller-reviewer: **[blocking]** substring version matching can produce a false-positive boundary

`bare_old in first_line` is a raw substring test. If `old_version`'s numeric prefix happens to be a substring of a *newer* version in the changelog (e.g. `old="1.2"` vs. a changelog entry versioned `1.20.3-1`), the cut lands on the wrong entry. Verified locally: with the newest entry at `1.20.3-1` and `old="1.2"`, the function returns an **empty** trimmed changelog while reporting `boundary_found: True` — silently discarding real newer entries and reporting false confidence, which is worse than the documented "no match found, keep everything" fallback.

Suggested fix: anchor the match to the version token boundary, e.g. `re.search(rf"-\s*{re.escape(bare_old)}\s*$", first_line)`.

plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
```diff
@@ -2543,10 +2561,115 @@ def _collect_rpmdb(
...
+    ) -> Optional[dict]:
+        """Extract and trim one package's changelog from the target rpmdb."""
+        rel_path = f"{target_tag}/rpmdb/{variant}/changelogs/{pkg}.txt"
+        output_path = os.path.join(variant_dir, "changelogs", f"{pkg}.txt")
+        if os.path.exists(output_path):
+            _log(f"    {variant}/{pkg}: skip (exists)")
+            return {
+                "variant": variant, "package": pkg, "old": old, "new": new,
+                "changelog": rel_path,
+            }
```
petr-muller-reviewer: **[should-fix]** `boundary_found` missing from cache-hit return value

The skip-existing-file path returns a dict without `boundary_found`, while the fresh-extraction path (further below, lines 2663-2666) includes it. SKILL.md/AGENTS.md document `boundary_found` as always present on `rpm_changelogs[]` entries — a re-run of the script (the designed skip path) will silently produce entries missing the field. Suggest persisting the value (it's already written into the `.txt` header) so the skip path can reconstruct it, or defaulting the key.

plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
```diff
@@ -2543,10 +2561,115 @@ def _collect_rpmdb(
...
+                variant_changes = cumulative.setdefault(variant, {})
+                for pkg, versions in (stream.get("changed") or {}).items():
+                    if pkg in variant_changes:
+                        variant_changes[pkg]["new"] = versions.get("new", "")
+                    else:
+                        variant_changes[pkg] = dict(versions)
+
+        if not cumulative:
+            return []
+        if not _check_rpm():
```
petr-muller-reviewer: **[nit]** `_check_rpm()` runs once in `main()` as a preflight warning, then again here after building the cumulative diff. Harmless (cheap call) but avoidable duplication.

.claude-plugin/marketplace.json
```diff
@@ -41,7 +41,7 @@
         "repo": "rh-uxd/ai-helpers",
         "path": "plugins/patternfly"
       },
-      "description": "PatternFly development tools from the UXD team — component scaffolding, design guidance, auditing, migration, test generation, and an MCP server.",
+      "description": "PatternFly development tools from the UXD team \u2014 component scaffolding, design guidance, auditing, migration, test generation, and an MCP server.",
```
petr-muller-reviewer, line 44: **[question]** unrelated diff hunk — re-escapes an em dash in the unrelated `patternfly` plugin's description. Likely a `make update` side effect on the whole file; worth confirming it's expected rather than an accidental revert of someone else's edit.

petr-muller-reviewer (changes requested):
This PR adds RPM changelog extraction for packages changed vs. baseline. The feature is well-integrated — it reuses already-fetched per-hop changelog data instead of a new API call, and the gating (`--no-rpm-changelogs`, `rpm` CLI preflight) mirrors the existing `podman`/`--no-rpmdb` pattern.

One correctness issue needs fixing before merge: the changelog-trimming heuristic (`_trim_changelog_to_new_entries`) uses substring matching to find the version boundary, which can produce a false-positive match and silently return an empty changelog while claiming `boundary_found: True` (see inline comment). A related schema-consistency issue exists on the cache-hit path.

Checked and fine:
- `subprocess.run` calls use argument lists (no shell) — no injection risk.
- Package names in filesystem paths / `rpm -q` args come from RPMDB-extracted data, not external input.
- `variant` naming is consistent between `_extract_rhcos_changes` and the RPMDB summaries being joined.
- The cumulative per-hop fold correctly reconstructs the target-vs-baseline diff (keeps first-seen `old`, updates `new` each hop).
- Version bumps in `plugin.json`/`marketplace.json` follow the project's contributing rule.
- Script compiles cleanly.

Open questions:
- Is the substring-matching approach in `_trim_changelog_to_new_entries` intentional as a "loose but simple" heuristic, or was the false-positive case not considered?
- Any reason not to add a unit test for the trimming function, given the existing test-file convention in this directory (`test_sippy_fallback.py`, `test_test_classification.py`)?

No tests were added for the new logic; given the bug found above, I'd suggest adding at least a test for `_trim_changelog_to_new_entries` covering adjacent/overlapping version numbers.

petr-muller:
@coderabbitai review

plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
petr-muller-reviewer: Resolved in 18ebf99b — boundary matching now anchors to the trailing version token (`(?:^|[\s-]){bare_old}\s*$`) instead of a free substring test, so `old="1.2"` no longer false-matches inside `1.20.3-1`. Verified locally.

plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
petr-muller-reviewer: Resolved in 18ebf99b — the cache-hit path now reads back a small header from the previously-written changelog file (via `_read_changelog_cache_header`) and returns `boundary_found` from it, also re-extracting if the cached old/new no longer match. Schema is now consistent between fresh and cached entries.

plugins/ci/skills/payload-snapshot/scripts/payload_snapshot.py
petr-muller-reviewer: Resolved in 18ebf99b — the redundant `_check_rpm()` call inside `_collect_rpm_changelogs` was removed; only the `main()` preflight check remains.

petr-muller:
The automated reviews requested a ton of corner case handling, and I believe it spiraled the complexity for really no good reason.

petr-muller:
/close 

Superseded by https://github.com/openshift-eng/ai-helpers/pull/677

PR #671, opened 2026-08-04
feat(ci): require mechanism verification and self-review before root-cause and fix claims in bulk-triage-regressions

mkowalski:
## Summary

Hardens `bulk-triage-regressions` with lessons from two real duty runs where the AI mis-diagnosed the mechanism behind a correctly-identified failure — and, per review feedback, generalizes the lessons instead of encoding individual cases, then applies the same restructure to the skill's pre-existing narrative content.

## Changes to the skill

**Case catalog (`references/case-notes.md`)** — reference file (moved to `references/` per the agent skills spec, per review) holding the full narratives of **13 real duty-run incidents** (after three eval-driven slimming/benchmark rounds; see below) previously narrated inline in the skill body. Six of those were told *twice* (once in a Phase 3/4 rule, once in Pitfalls) with the tellings already drifting apart; each now lives once, and every rule site carries a 1-line case pointer instead. The two new source cases from this PR were initially added here too, then **removed after the eval matrix (below) showed the generic rules catch them without the narratives** — they live on as eval cases instead.

**Phase 3 — three cross-cutting analysis principles** (most specific rules and Pitfalls are instances of these; the model falls back to the principle when no specific rule matches):

- **Messenger vs. owner**: the component that reports a failure is presumed the messenger until artifacts prove it is the producer
- **Population claims require population evidence**: "resolved/stopped/all/only" claims must enumerate the population they quantify over
- **Consult known-cause catalogs before deriving from scratch**: symptom catalog, triage records, component-scoped JIRA, repo merge history, closed siblings

**Phase 3 — new mechanism rules** (from the two new cases, compressed to principles):

- Race/ordering theories require timestamp proof; measure the failure window's duration
- Bad-value failures: trace the value to its producer across the data flow
- Decode suspicious constants before writing "malformed"/"overflow"
- The claimed mechanism must predict the observed frequency
- Differential recovery check: the component that failed to recover owns the bug
- Read the owning repo's source before claiming a missing safeguard

**Phase 3 — mandatory mechanism self-review checklist** (the generalization; addresses the overfitting concern raised in review):

1. Enumerate the load-bearing assumptions in the mechanism claim; each needs an artifact citation or is downgraded to an explicit "unverified: possibly …" hypothesis
2. Check the theory's predictions against everything observed: frequency, window duration, recovery asymmetry, variant spread
3. Argue the strongest alternative mechanism and cite the evidence that discriminates; if nothing discriminates, confidence is not HIGH

**Phase 4 (act):**

- Suggested fixes carry a higher evidentiary bar than attribution, must be proportionate to the measured window, must never lead with weakening a safety mechanism, and triage duty never opens fix PRs

**Pitfalls section** — restructured from a mix of principles and long "Real example:" narratives into principles with case pointers; operational rules (API payloads, batch limits, URL conversion, artifact paths) stay inline unchanged.

## The two source cases (now eval cases 001/002)

- **OCPBUGS-100316**: the AI identified the terminal blocker (a VAP `paramKind` denial) but mis-diagnosed a ~2-second apiserver informer-warmup transient as a CRD-install-ordering race, and opened a `failurePolicy: Ignore` PR ([cluster-capi-operator#640](https://github.com/openshift/cluster-capi-operator/pull/640), closed). The CRD was `Established` 20s before the first denial; the real product bug was the vSphere actuator permanently losing its vCenter task ID after one failed status patch (masters recovered from the same denials; workers never did).
- **OCPBUGS-99918**: "CRI-O pinns computes a malformed GID mapping" was filed against CRI-O for `Cannot write gid mappings: 0-4294901760-65536`; pinns/CRI-O were pass-throughs and the producer was kubelet's user-namespace ID allocator randomly picking the top 65536-ID block of the 32-bit space (fixed upstream in [kubernetes/kubernetes#140190](https://github.com/kubernetes/kubernetes/pull/140190)). The filed deterministic-bug theory predicted every-pod failure, contradicting the observed n=1.

## Eval suite and baseline-variant matrix (was: follow-up)

Per review, this PR now ships an agent-eval-harness eval (`plugins/ci/evals/eval-bulk-triage-regressions.yaml`) with **5 cases replaying real mis-diagnosed duty-run incidents** as frozen Sippy regression snapshots (triage state scrubbed back to duty time) with live GCS artifacts:

| Case | Regression | Replays |
|---|---|---|
| `case-001` | 45337 | OCPBUGS-100316 VAP denial: race-vs-warmup, recovery asymmetry, fix proportionality |
| `case-002` | 44906 | OCPBUGS-99918 GID mapping: producer tracing, constant decoding, frequency prediction |
| `case-003` | 41888 | etcd-vs-router victim attribution (`clusteroperators.json` end-state oracle) |
| `case-004` | 45377 | SecurityPenetration interference: source gating, `test` triage required |
| `case-005` | 42994 | CBO ~1 Hz flap: permafail-vs-"self-resolved", duplicate avoidance (must find OCPBUGS-99220) |

Judges: 6 deterministic (output schema, **no-writes guardrail**, anti-lumping bucket minimum, owning component/triage type/classification, banned-fix patterns, existing-JIRA discovery) + 2 LLM rubrics (mechanism accuracy vs owning-team ground truth, evidence discipline).

**Matrix results** (1 run per cell, deterministic judges, opencode/claude runner):

| Variant | Skill content | Score | Notes |
|---|---|---|---|
| baseline (pre-PR main) | none of this PR | 29/30 | `case-002` owner = `Node / CRI-O` — **reproduces the historically-filed wrong attribution**; `case-001` mechanism is ordering-flavored with no persistent-defect finding |
| methodology | principles + self-review checklist, **cases 1–2 narratives and citations stripped** | 30/30 | correct kubelet allocator attribution incl. `2^32` boundary decode and ~1/65535 frequency argument; actuator non-recovery identified as primary defect |
| full catalog | everything | 30/30 | no measurable delta over methodology on these cases |

Conclusion drawn in this PR: the **generalized methodology carries the improvement**, so the Case 1/2 narratives were dropped from the catalog (they add prompt tokens without measurable benefit) and preserved as eval ground truth instead.

A **second benchmark round** applied the same test to the narratives covered by eval cases 003–005: a variant without Cases 3/4/6/7/10 scored 29/30, with the single degradation being the CBO-flap bucket classified "resolved" instead of "permafail" — exactly Case 4's lesson. Restoring only Case 4 flipped it back to permafail. Result: **four narratives dropped** (etcd-vs-router, CBO duplicate, SecurityPenetration gating, wrapper cause-timeline — ground truth now enforced by eval cases 003–005), **the ~1 Hz flap narrative kept** — the eval showed it still pays for its tokens. The remaining narratives keep their place: they have no eval coverage yet (several need harness-level instrumentation), so dropping them would be un-evidenced. Case numbering in `references/case-notes.md` was made contiguous (1–13) after the slimming, so the numbers above refer to incidents, not to current case IDs.

A **third, per-model benchmark round** ran the slimmed catalog against fable-5 (30/30), opus-5 (29/30) and the target model **opus-4.6 (27/30)** — opus-4.6 regressed to the historical CRI-O mis-attribution on the GID case and lumped cause windows, while a control run against the full catalog fixed the attribution. Two narratives (GID producer trace, wrapper cause-timeline; now Cases 12–13) were therefore **restored as demonstrably load-bearing for the target model**; verification: opus-4.6 scores 29/30 on the final catalog, and its one residual miss (duplicate-bug discovery) also fails with the full catalog, so it is not narrative-fixable. Net: catalog slimmed 17 → 13 narratives, every removal and every retention backed by a benchmark run. Caveats: n=1 per cell, LLM judges not yet run (needs judge-model credentials), runner was opencode rather than the harness's claude-code path.

### Review follow-ups

A self-review of the eval found and fixed two scoring holes and three over-strong skill rules:

- `banned_fix_absent` exempted any sentence containing a negation word, so a verdict proposing *"Set failurePolicy: Ignore … so it does not block machine patches"* scored clean. Negation now has to govern the match.
- `primary_bucket_correct` took the first regex-matching bucket, making the result depend on bucket emission order (which is not part of the output contract). It now evaluates every matching bucket.
- `no_unauthorized_writes` matched bare tool names, so simply opening the `triage-regression` or `manage-symptoms` skill failed a read-only run. Name-only patterns removed; the authoritative control is that the eval environment carries no Sippy/JIRA credentials.
- Phase 3 rules no longer over-claim: an earlier readiness timestamp refutes only that ordering theory (not the whole timing angle); survivors show the failure was not uniformly permanent rather than proving a transient trigger; and fix proportionality no longer forbids a permanent *resilience* fix for a short trigger window — which is precisely the correct fix in case-001.

Re-scoring all ten recorded runs after these changes reproduces the pass rates above, so no benchmark conclusion changed.

A second review round addressed the four remaining CodeRabbit findings. Three were adopted: the producer of a bad value is now the *prime suspect* rather than the automatic owner (confirm it was wrong as produced — no intermediate hop corrupted it, no contract the consumer had to enforce); self-review accepts any specific checkable evidence locator instead of demanding a line number that run metadata and API fields do not have; and the Sippy host split is documented after verifying it — the symptom/label catalog is **public and unauthenticated** on `sippy.dptools.openshift.org`, only `reevaluate` is token-gated on `sippy-auth` (404 vs 403 on the public host), so a 401/403 is an unknown and *never* "no symptoms matched", with a credential-free fallback ladder.

The fourth — rewording the frequency rule so a differing precondition can explain n=1 — was **declined on eval evidence**. Verifying it on the target model showed [case-005](https://redhat.atlassian.net/browse/case-005) stopped splitting its two cause windows. Bisect, three runs per variant ([case-005](https://redhat.atlassian.net/browse/case-005) bucket split, need ≥2):

| Variant | Splits |
|---|---|
| pre-fix | 3/3 |
| all four edits | **0/4** |
| frequency edit reverted, auth text verbose | 1/3 |
| auth text compressed + frequency edit guarded | 2/3 |
| frequency edit reframed as a homogeneity check | 1/2 |
| **three edits kept, frequency edit reverted** | **3/3** |

Every wording that explains a rate by a differing precondition also licenses merging heterogeneous runs into one mechanism "firing under varying conditions" — precisely the lumping the cause-timeline rule prevents. The auth fix was also compressed: Phase 3 keeps one line, the ladder moved to the Authentication section, since the bisect showed prose volume in the analytical hot path matters too.

Final verification: a full five-case run on `claude-opus-4-6` against the merged state scores **30/30, zero regressions** (and found OCPBUGS-99220, which the earlier baseline missed).

**Known model gap (not a skill defect):** claude-opus-4-6 does not surface OCPBUGS-99220 on `case-005`, and misses it identically with the full pre-slimming catalog — so no narrative or rule fixes it; fable-5 and opus-5 find it. `expected_jira_found` is therefore set to `min_pass_rate: 0.8` with the reason recorded in the case annotations, rather than shipping a suite that lands permanently red on the target model. The other four cases remain hard gates.

Version bumped to 0.0.82; `make update` and `make lint` (A+) run.

Rebased on main after #672 merged, so the `docs/index.html` diff here is just the ci version bump.

Note: independent of #663 (terminal-blocker timeline check), which came from the same duty run — that PR covers finding the terminal blocker; this one covers verifying the mechanism behind it.

---
_This PR description was generated using AI. Please verify before acting on it._

## Summary by CodeRabbit

* **Updates**
  * Updated the CI plugin to version 0.0.81.

* **Documentation**
  * Strengthened regression triage guidance with stale-case checks, evidence validation, ownership analysis, and adversarial review.
  * Added 15 documented triage cases covering common attribution, monitoring, rollout, connectivity, and process failures.


theobarberbany:
This  feels like over fitting to me - the more generalised case is that we messed up the prelim investigation on the codebase :( 

What we’ve done above is potentially reject legit failures now because we saw this failure mode once. 

I’d be far more interested (and think the skill would improve more across different cases) if we could: 

- Add an eval for this case

- Iterate on the command (the harness supports this) to improve the investigation side. 

My intuition here says we want to get the model to review it’s work / suggestion and challenge any unfounded assumptions it can’t evidence.

mkowalski:
> This feels like over fitting to me - the more generalised case is that we messed up the prelim investigation on the codebase :(

Yes and no. This skill intentionally has the "generic" section and "examples" section.

I don't know yet how it's going to move forward in the future, but till now I have found out that having dozens of specific examples what we did wrong in the past is helping it. The reason (at least what I believe in) is that we are all the time analysing stuff from the same areas and our space of possible issues is quite limited.

Plus, we need to be good enough to route "installation should succeed but failed" to e.g. "CRI-O issue" but we don't (yet) aim for fully autonomous mode.

In the ideal world you would receive an investigation with a correct PR you just need to approve.

Intermediate step is that you are receiving an investigation with 50% wrong stuff, but it's still you who is responsible.

Without this step at all, no one is looking at the issue at all, because there was no single error message pointing towards your team.

theobarberbany:
> I have found out that having dozens of specific examples what we did wrong in the past is helping it. The reason (at least what I believe in) is that we are all the time analysing stuff from the same areas and our space of possible issues is quite limited.

Fair point that domain-specific examples help, the problem space is constrained and a model that's seen "race theory was wrong because timestamps said otherwise" will be more suspicious next time. The question is where those examples should live.
 
Right now they're baked into the prompt as rules. That means they cost tokens on every invocation (including the 95% of cases where the failure isn't a race theory or a bad-value trace), you can't measure whether they actually help (did the model get the next case right because of the rule, or because it was already doing better investigation?), and you can't iterate, if a rule causes false negatives you won't know until someone catches it manually.

This skill is running on Fable, so prompt size has a direct cost multiplier (and a big one). Every case-specific rule that doesn't pull its weight is burning money on every invocation.

My argument is more, you need evals for a command like this running in an automated fashion.  The actual cost of OCPBUGS-100316: one engineer took the AI output at face value and raised a PR that would have been catastrophic to merge. Another engineer reviewed it and escalated to the SME (me). I then spent a day  investigating the PR, working out why the AI wrote the ticket in the first place, and writing up the correct analysis. That's ~3 days of engineering time across three people, plus the trust cost, receiving confidently-wrong AI slop that would have broken production if merged is the kind of thing that makes teams stop engaging with AI triage entirely. False positives matter **massively** here.

With an eval for this case (and more for the skill in general) you get a tighter loop: run the skill against the OCPBUGS-100316 artifacts, check whether it produces the wrong diagnosis, tweak the investigation methodology (not the case-specific rules), re-run, repeat. If the general methodology handles it, the case-specific rules are dead weight. If it doesn't after several iterations, then you've earned the specific rule and you know exactly which one made the difference, not a bundle of six. You might even be able to prove it doesn't need fable / expensive models most of the time..

@stbenjam has done some awesome work on evals for the payload analysis commands,  which you could directly borrow from: https://github.com/openshift-eng/ai-helpers/tree/main/plugins/ci/evals  see also #wg-agent-eval-harness 

The mechanism self-review section (enumerate assumptions, check predictions, argue alternatives) is the one piece I'd keep from this PR regardless: that's methodology improvement, not pattern-matching. It's the kind of thing that would have caught both cases without needing the case-specific rules, and which you can *prove* with evals.

plugins/ci/skills/bulk-triage-regressions/references/case-notes.md
stbenjam, line 1: This should go a references/ dir per the agent skills spec

mkowalski:
> The actual cost of OCPBUGS-100316: one engineer took the AI output at face value and raised a PR that would have been catastrophic to merge. Another engineer reviewed it and escalated to the SME (me). I then spent a day investigating the PR, working out why the AI wrote the ticket in the first place, and writing up the correct analysis. That's ~3 days of engineering time across three people, plus the trust cost, receiving confidently-wrong AI slop that would have broken production if merged is the kind of thing that makes teams stop engaging with AI triage entirely. False positives matter massively here.

This particular discussion I keep having on regular basis. It's 3 days of engineering across three people, but your starting point is already with the issue somehow split into smaller pieces and _some_ investigation performed.

A year ago no one would ever look into this bug because how it exposes itself is a very generic installation failure (so the error you see is literally `install should succeed but did not`). In fully manual mode, no one ever looks into those. The failure is too generic and we don't have people with skills wide enough to look at every single installation failure to figure out whether today it failed because of networking, storage, machinery or maybe just infra failure. So they just keep hanging there, sometimes it's a bug that will never happen in the wild but some other cases are causing disruptions that we discuss for hours and we have no clue what is going on.

From this perspective, we have the following choices

A) do nothing; let it be; maybe a customer hits it, maybe not
B) have the AI do wrong investigation, we end up with 3 man-days and get the bug fixed
C) have the AI do the correct investigation and craft a fix; we end up with 0 man-days

C is ideal. Today we are at B. A is where we have been for the past X years.

The thing I agree 10/10 is the loss of trust, this is something hard to quantify because on one hand I prefer the AI to throw you a bug which is wrong but at least you are a person from the area; on the other hand whenever I get a bug with incorrect analysis, as an engineer I am annoyed. But from the perspective of someone who wants to get rid of the "mysterious failures", it is a success that it's you who got the bug and not someone from the storage team.

Anyway, it's not something we will ever solve for good. When we had support engineers (or anyone else, you name it) doing the triage/analysis we always had incredibly long ping-pongs between the teams ("it's not me", "it's not me either, maybe it's him", "no it's also not me, it's the first guy", [...])

theobarberbany:
> From this perspective, we have the following choices
A) do nothing; let it be; maybe a customer hits it, maybe not
B) have the AI do wrong investigation, we end up with 3 man-days and get the bug fixed
C) have the AI do the correct investigation and craft a fix; we end up with 0 man-days
C is ideal. Today we are at B. A is where we have been for the past X years.



I'm not arguing that we should not aim for B or even C,  I think where we are and what work we're doing is great in this area. 

I'm asking you to prove that the changes in this PR actually impact the issues we've seen with OCPBUGS-100316, and the way to do that is with evals. 

We can go B -> C faster with evals, than vibe coding prompts and praying we don't introduce regressions.

mkowalski:
/hold

I am working towards removing cases from narratives for the ones that we don't need. Running evals with and without those and the ones that don't contribute to the quality, will get removed.

mkowalski:
Evals will run via https://github.com/openshift/release/pull/83018

mkowalski:
/hold cancel

Ready to review

PR #679, opened 2026-08-05
TRT-2887: Significantly improve `analyze-disruption` skill

smg247:
Add `find_disruption_runs.py` for resolving Grafana disruption dashboard URLs into Prow job runs via Sippy API, with auto-selection of diverse representative samples (dedup, tier-based selection, job diversity, clean comparison from same job).

Add `download_timelines.py` for parallel artifact downloading across multiple runs.

Enhance `parse_disruption.py` with --format summary (one-line triage), --blast-radius (other disrupted backends), E2E test name extraction for cross-run correlation, and format_blast_radius for compact multi-backend output.

Restructure SKILL.md: Grafana URL input flow, summary-first workflow to avoid context overflow, clean comparison A/B analysis guidance, simplified Jira search, collapsed redundant steps.

Includes 55 tests across all three scripts.

## Which issue(s) this PR fixes:

Fixes 

## Checklist:
- [ ] Subject and description added to both, commit and PR.
- [ ] This change includes docs.

## Summary by CodeRabbit

* **New Features**
  * Disruption analysis now accepts Grafana dashboard URLs alongside Prow job URLs.
  * Added automatic run discovery, timeline collection, representative-run selection, and clean-run comparisons.
  * Enhanced reports with summaries, E2E results, blast-radius analysis, dashboard context, and Jira guidance.
  * Added a `--skip-jira` option.

* **Changes**
  * Updated the CI plugin to version 0.0.81.
  * Expanded command documentation for Grafana inputs.

* **Tests**
  * Added comprehensive coverage for run discovery, downloads, parsing, reporting, and error handling.


plugins/ci/skills/analyze-disruption/find_disruption_runs.py
```diff
@@ -0,0 +1,568 @@
...
+            "disruption_seconds": row.get("disruption_seconds", 0),
+        })
+    return lookup
+
+
+def max_disruption_for_backend(disruption_entries, base_backend):
+    """Find the max disruption seconds matching the base backend name."""
+    if not disruption_entries:
+        return None
+    matching = [e["disruption_seconds"] for e in disruption_entries if base_backend in e["backend_name"]]
```
mstaeble: `base_backend in e["backend_name"]` is a substring test, which conflates cache and non-cache backends: for a target like `"kube-api"`, this also matches `"cache-kube-api-new-connections"` since `"kube-api"` is a substring of it.

Repro:
```
    base_backend, _ = parse_backend("kube-api-new-connections")  # -> "kube-api"
    entries = [
        {"backend_name": "cache-kube-api-new-connections", "disruption_seconds": 300},
        {"backend_name": "kube-api-new-connections", "disruption_seconds": 2},
    ]
    max_disruption_for_backend(entries, base_backend)  # returns 300, not 2
```

Since the skill's own backend-classification guide treats cache vs non-cache disruption as pointing to different root causes, this means the "Disruption (s)" column and `select_representative_runs`'s tiering can rank/select runs by an unrelated backend's severity instead of the one the user is actually investigating from the Grafana dashboard. Suggest matching the exact backend name (respecting the cache prefix) rather than substring containment.

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
```diff
@@ -0,0 +1,568 @@
...
+    backend = args.backend or grafana_params.get("backend")
+
+    if not release:
+        print("Error: --release is required (or provide --grafana-url with var-releases)", file=sys.stderr)
+        sys.exit(1)
+    if not backend:
+        print("Error: --backend is required (or provide --grafana-url with var-backend)", file=sys.stderr)
+        sys.exit(1)
+
+    base_backend, _conn_type = parse_backend(backend)
```
mstaeble, line 514: `_conn_type` is parsed but discarded here, so `max_disruption_for_backend` also merges `-new-connections` and `-reused-connections` readings for the same base name. Combined with the cache/non-cache substring issue on line 163, this means the reported disruption seconds can reflect up to 4 distinct backends (cache/non-cache × new/reused) instead of the one specified in `var-backend`. Worth threading the connection type through so matching is exact.

plugins/ci/skills/analyze-disruption/test_find_disruption_runs.py
```diff
@@ -0,0 +1,402 @@
...
+    assert result == ["metrics-api"]
+
+
+def test_max_disruption_for_backend_matching():
+    entries = [
+        {"backend_name": "kube-api-new-connections", "disruption_seconds": 73},
+        {"backend_name": "cache-kube-api-new-connections", "disruption_seconds": 75},
+        {"backend_name": "oauth-api-new-connections", "disruption_seconds": 5},
+    ]
+    assert max_disruption_for_backend(entries, "kube-api") == 75
```
mstaeble: This assertion locks in the substring-conflation bug from `find_disruption_runs.py:163` rather than catching it — `75` is the disruption value from the *cache* variant (`cache-kube-api-new-connections`), not from the `kube-api-new-connections` entry (`73`) being queried. Once the matching logic is fixed to be exact, this test should assert `73` and probably add a case confirming the cache entry is excluded.

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
```diff
@@ -0,0 +1,568 @@
...
+    since_ms = int((time.time() - args.since_hours * 3600) * 1000)
+    filter_dict = build_sippy_filter(variants, since_ms)
+    rows = fetch_runs(release, filter_dict, args.limit)
+
+    if not rows:
+        filters_desc = ", ".join("%s:%s" % (k, v) for k, v in variants.items())
+        print("No runs found for release=%s with variants [%s] in the last %d hours." % (
+            release, filters_desc, int(args.since_hours)), file=sys.stderr)
+        print("Try widening --since-hours or relaxing filters.", file=sys.stderr)
+        sys.exit(0)
```
mstaeble, line 539: Nit: this "no runs found" path exits `0` while the missing `--release`/`--backend` checks above exit `1`. Since this is a legitimate no-op-but-not-quite-success case (and is reported via a message on stderr), consider whether it should exit non-zero for consistency, or leave a short comment noting the exit(0) is intentional for interactive callers that parse stdout/stderr rather than exit codes. (Same applies to line 543.)

plugins/ci/skills/analyze-disruption/parse_disruption.py
mstaeble, line 801: `summarize_e2e_tests` (added above) returns `{"count", "failed_tests", "passed_tests", "tests"}` — no `"events"` key. Since `E2ETest` isn't special-cased like `OVSVswitchdLog`/`CPUMonitor`/`CloudMetrics` above, it falls into this generic branch, and `detail.get("events", [])` is always `[]`. So `--format text` only ever prints `E2ETest: N events` with no test names.

Repro: build a data dict with an E2ETest concurrent event named `"suspicious-test-that-causes-pressure"` and run it through `format_text()` — the test name never appears in the output.

Since E2E-test correlation is one of this PR's headline features and SKILL.md Step 3.3 recommends `--format text` for most deep-dives (reserving `--format json` for when text is "insufficient"), this signal is effectively invisible unless the caller specifically knows to use `--format json`. Suggest adding an `E2ETest` branch here (listing `failed_tests`/`passed_tests`), and ideally surfacing it in `format_summary` too.

plugins/ci/skills/analyze-disruption/download_timelines.py
```diff
@@ -0,0 +1,206 @@
...
+        "error": None,
+    }
+
+    logs_dir = os.path.join(output_dir, build_id, "logs")
+    os.makedirs(logs_dir, exist_ok=True)
+
+    prowjob_local = os.path.join(logs_dir, "prowjob.json")
+    prowjob_gcs = "gs://%s/logs/%s/%s/prowjob.json" % (GCS_BUCKET, job, build_id)
+
+    if not download_file(prowjob_gcs, prowjob_local):
```
mstaeble, line 115: `process_run` always re-downloads `prowjob.json` and every timeline file, even if they already exist locally from a prior invocation — there's no `os.path.exists(local_path)` short-circuit here. SKILL.md's Step 2 tells the *agent* to check for existing artifacts and ask before re-downloading, but that's enforced only by the agent following instructions, not by the script. If the flow gets re-run (e.g. the user changes their run selection and the agent re-invokes this script for overlapping runs), it'll silently redo all the `gcloud storage ls`/`cp` calls for data already on disk. Worth adding a skip-if-exists check here for defense in depth.

plugins/ci/skills/analyze-disruption/SKILL.md
```diff
@@ -67,152 +88,244 @@ The user will provide:
...
 
-```bash
-# Step 1: List all timeline files in the job's artifact tree
-gcloud storage ls "gs://test-platform-results/logs/{job_name}/{build_id}/artifacts/**/e2e-timelines_spyglass_*.json"
+Use `--disruption-only` to filter to runs with disruption > 0 for the target backend.
+This uses actual BigQuery data, not just test failures.
+
+Additional flags:
+- `--since-hours N` — change lookback window (default: 720 = 30 days)
+- `--limit N` — max runs to fetch (default: 50)
```
mstaeble, line 170: Nit: `--limit` defaults to 50 and `--since-hours` to 720 (30 days), so every Grafana-URL resolution pulls BigQuery disruption data for up to 50 runs even when `--auto-select` only needs 5. That's probably necessary for the tiering/diversity algorithm to have a large enough candidate pool, but it might be worth a line in the docs suggesting users narrow `--limit`/`--since-hours` for tighter dashboards where a smaller candidate pool is fine — gives an explicit cost lever rather than always defaulting to the widest query.

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
smg247: exit(0) is correct here — this isn't an error, it's a valid empty result. The agent parses stderr messages, not exit codes.

plugins/ci/skills/analyze-disruption/download_timelines.py
smg247: Defense in depth, but the SKILL.md already instructs the agent to check. Adding filesystem checks adds complexity for a case that's already handled at the agent layer.

I have also tried this out manually, and the agent skips downloading files it already has.

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
mstaeble: This overcorrects: it excludes `cache-` backends unconditionally, even when the *target itself* is a cache backend.

    base, _ = parse_backend("cache-kube-api-new-connections")   # -> "cache-kube-api"
    entries = [{"backend_name": "cache-kube-api-new-connections", "disruption_seconds": 300}]
    max_disruption_for_backend(entries, base)   # returns 0, not 300

Cache backends (`cache-kube-api`, `cache-oauth-api`, `cache-openshift-api`) are one of the two primary backend buckets this skill is built around, so a Grafana dashboard targeting a cache backend would now show 0s disruption for every run, and `select_representative_runs` would find zero candidates. The new test (`test_max_disruption_for_backend_matching`) only covers the non-cache-target case, so this isn't caught.

Suggest matching cache-ness symmetrically instead of excluding it outright:

    def _backend_matches(backend_name, base_backend):
        if backend_name.startswith("cache-") != base_backend.startswith("cache-"):
            return False
        return backend_name == base_backend or backend_name.startswith(base_backend + "-")

Worth adding a test case where `base_backend` is itself a cache variant.

PR #682, opened 2026-08-06
fix node-cve: scope triage to latest OCP version by default

Chandan9112:

Add OCP version filtering to `node-cve:triage` so it only triages CVE trackers for the latest OCP version by default (currently 5.0). Older versions are owned by the OCP sustaining team and should not receive Node-team triage comments.

Previously, `/node-cve:triage --notify-jira` would triage and post analysis comments to tracker issues across **all** OCP versions (4.12.z through 5.0). The sustaining team owns triage for all versions except the latest in-development release, so this created noise and duplicated their work.

This PR adds a `--version` flag with three modes:
- **`latest` (default):** auto-detect the highest OCP version from the Jira query results and triage only that version's trackers. No version is hardcoded — the detection is dynamic and advances automatically as new OCP releases ship.
- **A specific version** (e.g., `--version 5.0`): triage only that version.
- **`all`**: triage all versions (legacy behavior, for cross-version analysis when needed).

The implementation follows the same defense-in-depth pattern established in PR #623 (component safeguard):

**`query-open-cves`:**
- Adds Step 5 (Filter by OCP version) after deduplication. Versions are parsed as `(major, minor)`, `.z` suffixes are stripped for comparison, and the highest version is auto-detected. CVEs with no tracker for the target version are excluded entirely.

**`report-findings`:**
- Adds **OCP Version Safeguard** section — a posting-time re-validation of each tracker's OCP version against the active version filter, independent of Phase 1's filtering. This is the same defense-in-depth principle as the component safeguard.
- Audit log now separately tracks version-mismatch skip counts alongside component-mismatch skips.
- Validation pseudocode uses a single `jira issue view` call per tracker (fetching COMPONENT and SUMMARY in one request) to minimize API usage and rate-limiting risk.

**`analyze-cve-repos`:**
- Notes that input is version-filtered from Phase 1, and explains behavior under both `--version latest` and `--version all`.

**`triage.md`:**
- Adds `--version` to argument parsing, Phase 1 steps, clone strategy, Phase 3 posting, Phase 4 summary output, examples, and notes.

Also bumps `node-cve` to `0.0.7` and syncs marketplace/docs metadata.

## Summary by CodeRabbit

- **New Features**
  - Triage now automatically targets the latest detected OCP version and its release branch.
  - Reports and command output include version scope, filtering details, cache status, and posting audits.
  - CVE results now include version-aware tracker totals and metadata.

- **Bug Fixes**
  - Older-version trackers and branches are excluded from analysis.
  - Invalid versions, missing branches, and mismatched trackers are safely skipped and logged.
  - Added validation to prevent inappropriate Jira updates.

- **Documentation**
  - Updated plugin guidance, examples, and workflow documentation.
  - Updated the plugin version to 0.0.7.

### DIFF
--- a/plugins/node-cve/README.md
+++ b/plugins/node-cve/README.md
@@ -27,7 +27,7 @@ Triage all open CVEs for Node team components with automated reachability analys
 
 1. Queries OCPBUGS for open Vulnerability issues across all Node team components (CRI-O, Kubelet, MCO, etc.)
 2. Deduplicates by CVE ID (each CVE has multiple version trackers)
-3. Filters to the target OCP version (default: latest — currently 5.0). Older versions are owned by the sustaining team.
+3. Filters to the target OCP version (default: latest — auto-detected as the highest numeric version from query results, e.g., `5.0`). Older versions are owned by the sustaining team.
 4. Clones affected repositories at the target version's release branch and analyzes source code for reachability
 5. Classifies each CVE: Reachable, Present but not exploitable, Present but not reachable, Unaffected, or Uncertain
 6. Generates a triage report with confidence levels and recommended actions
@@ -104,7 +104,7 @@ spec:
 
 ## Safeguards
 
-The plugin implements three layers of defense to ensure triage analysis is only posted where it belongs:
+The plugin implements two safeguards, each with query-time and posting-time validation, to ensure triage analysis is only posted where it belongs:
 
 ### Component safeguard (cross-team protection)
 
--- a/plugins/node-cve/README.md
+++ b/plugins/node-cve/README.md
@@ -14,7 +14,7 @@ Requires the `node-team` plugin (installed automatically as a dependency).
 
 ## Command
 
-### `/node-cve:triage [--component <name>] [--notify-jira] [--notify-slack] [--days N]`
+### `/node-cve:triage [--component <name>] [--version latest|<ver>|all] [--notify-jira] [--notify-slack] [--days N]`
 
 Triage all open CVEs for Node team components with automated reachability analysis.
 
@@ -27,14 +27,16 @@ Triage all open CVEs for Node team components with automated reachability analys
 
 1. Queries OCPBUGS for open Vulnerability issues across all Node team components (CRI-O, Kubelet, MCO, etc.)
 2. Deduplicates by CVE ID (each CVE has multiple version trackers)
-3. Clones affected repositories at all version-specific release branches and analyzes source code for reachability
-4. Classifies each CVE: Reachable, Present but not exploitable, Present but not reachable, Unaffected, or Uncertain
-5. Generates a triage report with confidence levels and recommended actions
-6. Posts analysis comments to Jira tracker issues (with `--notify-jira`)
-7. Sends a threaded summary to Slack (with `--notify-slack`)
+3. Filters to the target OCP version (default: latest — currently 5.0). Older versions are owned by the sustaining team.
+4. Clones affected repositories at the target version's release branch and analyzes source code for reachability
+5. Classifies each CVE: Reachable, Present but not exploitable, Present but not reachable, Unaffected, or Uncertain
+6. Generates a triage report with confidence levels and recommended actions
+7. Posts analysis comments to Jira tracker issues (with `--notify-jira`)
+8. Sends a threaded summary to Slack (with `--notify-slack`)
 
 **Arguments:**
 - `--component <name>`: Filter to a single specific component (e.g., "Node / CRI-O"). If omitted, ALL Node team components are included — and only Node team components, never all OCPBUGS components.
+- `--version <value>`: Control which OCP versions are triaged. Default: `latest` (auto-detects the highest OCP version from query results). Use a specific version (e.g., `5.0`) or `all` for full cross-version analysis.
 - `--notify-jira`: Post analysis results as comments on Jira tracker issues (also enables cross-run caching)
 - `--notify-slack`: Send summary to Slack (API token for threading, or webhook for simple messages)
 - `--days N`: Only include CVEs updated in the last N days (default: all open)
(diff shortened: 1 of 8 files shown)


plugins/node-cve/skills/query-open-cves/SKILL.md
```diff
@@ -72,7 +72,36 @@ A single CVE may span multiple components (e.g., both "Node / CRI-O" and "Machin
...
+2. Collect all OCP versions extracted from tracker summaries in Step 3 (e.g., `4.12.z`, `4.14.z`, `4.17`, `4.18`, `4.19`, `5.0`).
+3. **Discard non-numeric versions:** Any version string that cannot be parsed as a numeric `major.minor` pair (e.g., `latest`, `nightly`, or other non-version labels) must be discarded with a warning — do not treat it as a valid OCP version. Only versions matching the pattern `<digits>.<digits>` (optionally followed by `.z`) are valid. If ALL extracted versions are discarded (no valid numeric versions remain), treat this the same as zero Jira rows: return an empty CVE list with `version_filter: null`, `version_filter_mode: "latest"`, and `version_detected: false`. Print: "No valid numeric OCP versions found in tracker summaries; version auto-detection failed."
+4. Determine the latest version by sorting the remaining valid versions numerically:
+   - Strip trailing `.z` suffixes for comparison purposes (`.z` indicates a z-stream release, which is always an older maintenance stream).
```
saschagrunert, line 83: The `.z` stripping is specified for sorting, but the text never states what form `version_filter` stores. Point 5 (line 89) says "remove tracker issues whose OCP version does not match the latest version" without saying whether "the latest version" retains or strips the `.z`.

The return value example (`"version_filter": "5.0"`) doesn't disambiguate since `5.0` never had a `.z`. If the latest version is detected from a set like `[4.12.z, 4.13.z, 4.14.z]`, is `version_filter` `"4.14"` or `"4.14.z"`?

Downstream in `report-findings`, `version_matches_filter()` strips `.z` from the tracker version before comparison but doesn't say it strips from `version_filter` too. If `version_filter` stores `"4.14.z"`, comparing stripped tracker version `"4.14"` against unstripped filter `"4.14.z"` would fail.

Suggest adding after point 4: "`version_filter` always stores the stripped `major.minor` form (no `.z` suffix), since `.z` is a release-stream qualifier, not a distinct version."

plugins/node-cve/skills/query-open-cves/SKILL.md
```diff
@@ -95,14 +128,19 @@ Flag CVEs where:
...
+      "tracker_keys": ["OCPBUGS-85948"],
+      "affected_versions": ["5.0"],
       "labels": ["pscomponent:cri-o"]
     }
   ]
 }
 ```
 
+**Field semantics for `version_filter`:**
+- When a version is resolved (auto-detected or user-specified): `version_filter` is the resolved version string (e.g., `"5.0"`), `version_detected` is `true`.
```
saschagrunert: Two gaps in the field semantics:

1. **Missing `version_filter_mode` for `--version <specific>`:** This bullet covers both auto-detected and user-specified, but `version_filter_mode` is only defined for `"latest"` (line 141) and `"all"` (line 142). When the user passes `--version 5.0`, what is the mode? The triage command prints `"(user-specified)"` in the output, suggesting the mode should be something like `"specific"`. Suggest adding: "When `--version <specific>` is used: `version_filter_mode` is `"specific"`, `version_detected` is `true`."

2. **No concrete value for `version_detected` under `--version all`:** Line 142 says "not applicable" but the field will be serialized. Suggest either omitting the field or setting it to `true`.

plugins/node-cve/commands/triage.md
```diff
@@ -30,10 +30,16 @@ Designed for both interactive use and headless execution via `claude --print`.
 
 1. **Parse Arguments**
    - `--component <name>`: Filter to a single specific OCPBUGS component (e.g., "Node / CRI-O"). Optional. **If omitted, the command includes ALL Node team components — and only Node team components. It never queries or posts to components outside the Node team.** See the [node-team shared components reference](../../node-team/skills/node/references/shared/components.md) for the canonical list.
+   - `--version <value>`: Control which OCP versions are triaged. Default: `latest`. Accepted values:
```
saschagrunert: nit: No input validation is specified for invalid values. `--version foobar` would fall through to the "specific version" path in `query-open-cves` Step 5, silently match zero trackers, and exit with "No open CVEs for Node team components at version foobar."

The `--component` flag already validates against the shared reference and exits with an error on mismatch. Consider adding similar validation: if the value is not `latest`, `all`, or matching `<digits>.<digits>`, print an error and exit.

plugins/node-cve/commands/triage.md
```diff
@@ -269,22 +282,26 @@ See report-findings skill.)
...
 
-The headline format depends on whether cached results were used. On first run (all CVEs are new): "Node CVE Triage (N CVEs analyzed)". On subsequent runs with cached results: "Node CVE Triage (N CVEs, M new)" where M is the number of CVEs without a prior `node-cve:triage` Jira comment. If classifications changed since the last run, also show "K updated": "Node CVE Triage (N CVEs, M new, K updated)".
+The headline format depends on the version scope and whether cached results were used. On first run (all CVEs are new): "Node CVE Triage (N CVEs analyzed, OCP 5.0)". On subsequent runs with cached results: "Node CVE Triage (N CVEs, M new, OCP 5.0)". If classifications changed since the last run, also show "K updated". If `--version all` was used, replace the version with "all versions".
```
saschagrunert: The template on line 272 is `Node CVE Triage (N CVEs analyzed, OCP <version>)`. "Replace the version with 'all versions'" substitutes only the `<version>` placeholder, producing "OCP all versions". Should say "replace 'OCP <version>' with 'all versions'" to get "Node CVE Triage (N CVEs analyzed, all versions)".

saschagrunert (commented):
Good design overall. The dual-safeguard pattern (query-time + posting-time) mirrors the existing component safeguard cleanly. Four issues found, two of which could cause matching failures at runtime.

saschagrunert:
Based on the Slack conversation, the requirement is "only pick the latest version" since sustaining owns everything else. The `--version` flag (with `latest`/`<specific>`/`all` modes) adds significant complexity for use cases nobody has asked for yet:

- Three code paths in `query-open-cves` (auto-detect, specific, all)
- Version re-validation at posting time needs to handle all three modes
- Field semantics for `version_filter_mode` need to cover each case
- Input validation for arbitrary version strings

Suggest dropping `--version` entirely and hardcoding the "latest" behavior (auto-detect the highest OCP version from query results, filter to it). That matches what was discussed, and if someone needs cross-version analysis later, the flag can be added then.

This would also resolve most of the inline review comments, since they stem from the three-mode complexity.

plugins/node-cve/README.md
```diff
@@ -27,11 +27,12 @@ Triage all open CVEs for Node team components with automated reachability analys
 
 1. Queries OCPBUGS for open Vulnerability issues across all Node team components (CRI-O, Kubelet, MCO, etc.)
 2. Deduplicates by CVE ID (each CVE has multiple version trackers)
-3. Clones affected repositories at all version-specific release branches and analyzes source code for reachability
-4. Classifies each CVE: Reachable, Present but not exploitable, Present but not reachable, Unaffected, or Uncertain
-5. Generates a triage report with confidence levels and recommended actions
-6. Posts analysis comments to Jira tracker issues (with `--notify-jira`)
-7. Sends a threaded summary to Slack (with `--notify-slack`)
+3. Filters to the target OCP version (default: latest — auto-detected as the highest numeric version from query results, e.g., `5.0`). Older versions are owned by the sustaining team.
```
saschagrunert: "Filters to the target OCP version (default: latest ...)" still implies other modes exist. Since `--version` was dropped, suggest:

"Filters to the latest OCP version (auto-detected as the highest numeric version from query results, e.g., `5.0`). Older versions are owned by the sustaining team."

Same for "target version's release branch" on line 31.

plugins/node-cve/commands/triage.md
```diff
@@ -220,9 +224,9 @@ The per-branch results are preserved in the report so reviewers can see which ve
...
+   **Before posting anything, re-validate every tracker against both the auto-detected version and the Node team component list — do not rely solely on Phase 1's filtering.** See [report-findings](../skills/report-findings/SKILL.md) "Node Team Component Safeguard", "OCP Version Safeguard", and Step 2 for the mandatory validation logic and audit logging. Skip (never post to) any tracker whose component is not a Node team component or whose OCP version does not match the auto-detected latest version, and record it in the posting audit log.
 
    Use Atlassian wiki markup (not Markdown), matching the format in [report-findings](../skills/report-findings/SKILL.md) Step 2. The comment includes per-branch results across all analyzed versions so reviewers can see the full picture.
```
saschagrunert: "per-branch results across all analyzed versions" is stale. With latest-only there is one version and one branch. Remove or reword.

Also, the report template above (lines 212-221, outside this diff) still shows multi-version structures: `Affected versions | 4.12.z - 4.19` and a "Per-branch results" table with 3 branch rows. With latest-only these will always be a single version and a single row. Since these templates instruct Claude on what output to generate, updating them to single-version examples would avoid confusing output.

plugins/node-cve/skills/report-findings/SKILL.md
```diff
@@ -161,21 +188,23 @@ Search the output for comments containing `[node-cve:triage|`. This pattern anch
 - Post to ALL **validated** tracker issues for a CVE (all version trackers), each with its version-specific result
```
saschagrunert: "(all version trackers)" is stale. After latest-only filtering, each CVE has trackers for one version. Suggest: "Post to every **validated** tracker issue for a CVE."

Also, the Jira comment template above (lines 154-158, outside this diff) has `h4. Results across all analyzed branches` followed by a 3-branch example table. Same issue as the triage.md template: with latest-only there is one branch, so the heading and multi-row example are misleading.

saschagrunert (commented):
The simplification in bf9b37f resolved the complexity issues from the last round. Three stale multi-version references remain after the `--version` removal, plus two output templates (outside the diff) that still show multi-branch examples despite the command now analyzing only one branch.

saschagrunert (approved):

PR #683, opened 2026-08-06
TRT-2887: remove analyze-disruption command to prevent model freelancing

smg247:
Follows up on https://github.com/openshift-eng/ai-helpers/pull/679

The verbose command file gave the model enough context to improvise with curl/WebFetch instead of loading and following the SKILL.md steps. Skills are slash-invokable so the command isn't needed.

## Summary by CodeRabbit

* **Documentation**
  * Simplified the `ci:analyze-disruption` command synopsis.
  * Clarified that the command loads the analysis skill and forwards provided arguments.
  * Updated supported inputs to include Prow job or Grafana URLs, optional backend selection, and JIRA skipping.
  * Refreshed CI plugin metadata to version 0.0.82.


plugins/ci/commands/analyze-disruption.md
stbenjam, line 1: I'd suggest just deleting this.

Skills are slash-invocable.

stbenjam: I've also noticed, at least in older versions of claude code, it doesn't do well when a command and skill are named the same -- it often doesn't see the skill, or ignores it

PR #686, opened 2026-08-07
TRT-2887: multi-value Grafana params and new filter mappings

smg247:
### Support multi-value Grafana URL params
Grafana dashboards can have repeated `var-` params (e.g. `var-platform=azure&var-platform=gcp`). Previously only the first value was used. Now multi-value params are detected and separate Sippy queries are issued per combination, with results merged, deduped by prow_id, sorted by timestamp descending, and capped at the global `--limit`. Cartesian expansion is bounded at 20 combinations.

### Map FeatureSet, NetworkStack, and OS Grafana filters
Added `featureset` → `FeatureSet`, `ipmode` → `NetworkStack`, and `os` → `OS` to `GRAFANA_TO_VARIANT`. Without these, Grafana URLs with `var-featureset=techpreview` would return runs from all feature sets instead of just techpreview jobs. Display-only Grafana params (`master_nodes_updated`, `percentile`, `lookback`, etc.) are silently ignored; unknown params emit a stderr warning.

### Fix symmetric cache-ness matching
`_backend_matches` previously excluded cache backends unconditionally. When the target itself was a cache backend (e.g. `cache-kube-api`), all matching entries returned 0 disruption. Now matching respects whether both sides are cache or non-cache.

## Test plan
- [x] All 34 tests pass in `test_find_disruption_runs.py`
- [x] `make lint` passes clean (A+)
- [x] Tested `/ci:analyze-disruption` with Grafana URL — skill loads and follows SKILL.md steps correctly
- [x] Multi-platform URL (`var-platform=azure&var-platform=gcp`) queries both platforms and merges results

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disruption-run analysis for dashboards with repeated filter parameters.
  * Results are gathered across selected filter values, deduplicated, sorted by newest timestamp, and limited appropriately.
  * Searches support up to 20 filter combinations while preserving existing single-value behavior.
  * Added warnings for unsupported dashboard parameters and validation for conflicting selections.

* **Tests**
  * Added coverage for repeated filters, platform expansion, deduplication, sorting, limits, validation, and additional filter types.

* **Chores**
  * Updated the CI plugin version to 0.0.83.

### DIFF
--- a/plugins/ci/skills/analyze-disruption/find_disruption_runs.py
+++ b/plugins/ci/skills/analyze-disruption/find_disruption_runs.py
@@ -31,6 +31,14 @@ GRAFANA_TO_VARIANT = {
     "topologies": "Topology",
     "networks": "Network",
     "upgrade_type": "Upgrade",
+    "featureset": "FeatureSet",
+    "ipmode": "NetworkStack",
+    "os": "OS",
+}
+
+GRAFANA_DISPLAY_ONLY = {
+    "master_nodes_updated", "percentile", "lookback", "min_disruption_regression",
+    "min_disruption_job_list", "min_relevance", "orgId",
 }
 
 
@@ -530,6 +538,11 @@ def main(argv=None):
         val = cli_overrides.get(gkey) or grafana_params.get(gkey)
         if val:
             variants[variant_key] = val
+
+    known_keys = set(GRAFANA_TO_VARIANT) | GRAFANA_DISPLAY_ONLY | {"releases", "backend", "_dashboard_name"}
+    for gkey in grafana_params:
+        if gkey not in known_keys:
+            print("Warning: Grafana parameter '%s' not mapped to a Sippy filter, ignoring" % gkey, file=sys.stderr)
 
     since_ms = int((time.time() - args.since_hours * 3600) * 1000)
 
--- a/plugins/ci/skills/analyze-disruption/find_disruption_runs.py
+++ b/plugins/ci/skills/analyze-disruption/find_disruption_runs.py
@@ -35,14 +35,18 @@ GRAFANA_TO_VARIANT = {
 
 
 def parse_grafana_url(url):
-    """Extract var-* parameters from a Grafana dashboard URL."""
+    """Extract var-* parameters from a Grafana dashboard URL.
+
+    Multi-value params (e.g. var-platform=azure&var-platform=gcp) are stored
+    as comma-joined strings so downstream code stays simple.
+    """
     parsed = urllib.parse.urlparse(url)
     params = urllib.parse.parse_qs(parsed.query)
     result = {}
     for key, values in params.items():
         if key.startswith("var-"):
             name = key[4:]
-            result[name] = values[0]
+            result[name] = ",".join(values) if len(values) > 1 else values[0]
 
     dashboard_name = ""
     path_parts = parsed.path.rstrip("/").split("/")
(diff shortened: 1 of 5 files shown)


plugins/ci/skills/analyze-disruption/find_disruption_runs.py
```diff
@@ -527,9 +539,41 @@ def main(argv=None):
...
+    known_keys = set(GRAFANA_TO_VARIANT) | GRAFANA_DISPLAY_ONLY | {"releases", "backend", "_dashboard_name"}
+    for gkey in grafana_params:
+        if gkey not in known_keys:
+            print("Warning: Grafana parameter '%s' not mapped to a Sippy filter, ignoring" % gkey, file=sys.stderr)
+
     since_ms = int((time.time() - args.since_hours * 3600) * 1000)
-    filter_dict = build_sippy_filter(variants, since_ms)
-    rows = fetch_runs(release, filter_dict, args.limit)
+
+    multi_keys = [(k, v.split(",")) for k, v in variants.items() if "," in v]
```
petr-muller-reviewer, line 556: `multi_keys` is built only from `variants.items()` (the GRAFANA_TO_VARIANT-mapped keys), so it never expands `backend` or `release`. But `parse_grafana_url` comma-joins *any* repeated `var-*` param (including `var-backend` and `var-releases`) into a single string.\n\nA URL with `var-backend=host-to-host-new-connections&var-backend=kube-api-new-connections` produces `backend = "host-to-host-new-connections,kube-api-new-connections"`. That string is never split — it goes straight into `parse_backend()` at line ~524, which strips the `-new-connections` suffix from the whole string, yielding `base_backend = "host-to-host-new-connections,kube-api"` (verified locally). No real backend matches this, so `max_disruption_for_backend` silently returns 0 for every row instead of erroring or querying both backends.\n\nSame problem for `var-releases`: `release = "5.0,5.1"` is passed verbatim as the Sippy `release` query param, which almost certainly yields zero rows — the tool then prints "No runs found ... in the last N hours", misattributing an API-format bug to absence of data.\n\nEither expand `backend`/`release` into the combo loop the same way as `variants`, or explicitly reject/warn on multi-value `var-backend`/`var-releases` since they are not part of the Sippy variant filter and cannot be queried as one string.

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
```diff
@@ -527,9 +539,41 @@ def main(argv=None):
...
+        for mk, mv in multi_keys:
+            variant_combos = [
+                dict(combo, **{mk: val}) for combo in variant_combos for val in mv
+            ]
+        max_combos = 20
+        if len(variant_combos) > max_combos:
+            print("Error: %d variant combinations exceeds limit of %d" % (
+                len(variant_combos), max_combos), file=sys.stderr)
+            sys.exit(1)
+        for combo in variant_combos:
```
petr-muller-reviewer, line 568: Minor: up to 20 variant combinations are queried against Sippy sequentially in a blocking loop; each `fetch_runs()` call can take up to its 60s timeout. Since these are independent, I/O-bound requests, a `ThreadPoolExecutor`-based fan-out would bound end-to-end latency to roughly one query instead of up to 20 sequential ones.

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
```diff
@@ -527,9 +539,41 @@ def main(argv=None):
...
+
     since_ms = int((time.time() - args.since_hours * 3600) * 1000)
-    filter_dict = build_sippy_filter(variants, since_ms)
-    rows = fetch_runs(release, filter_dict, args.limit)
+
+    multi_keys = [(k, v.split(",")) for k, v in variants.items() if "," in v]
+    if multi_keys:
+        seen_prow_ids = set()
+        rows = []
+        variant_combos = [{}]
```
petr-muller-reviewer: Minor: this hand-rolled accumulator reimplements `itertools.product` — `itertools.product(*[mv for _, mv in multi_keys])` zipped with the keys does the same thing in one call and is easier to recognize than re-deriving a bespoke Cartesian product.

petr-muller-reviewer (changes requested):
Automated review pass (code-review, medium effort). Two correctness bugs in the new multi-value handling, plus minor cleanup suggestions.

petr-muller (approved):
/hold

the Python looked fine to me so I'm tagging this for merge if you prefer, but my reviewer claude thinks there are some subtle bugs so I let it post a review - feel free to ignore

petr-muller:
/lgtm

~hrm no approval-means-lgtm here?~

wasn't github better in dynamically updating a PR in an opened tab withtout refresh :unamused:

plugins/ci/skills/analyze-disruption/find_disruption_runs.py
smg247: Valid optimization in theory, but in practice queries are fast (<1s) and the common case is 2-4 combos. Not worth the threading complexity right now.

petr-muller (approved):

PR #695, opened 2026-08-12
Add Codex payload analysis eval

not-stbenjam:
## What changed

- add a Codex/Luna `xhigh` payload-analysis eval configuration with deterministic and Opus judges
- expose the full CI plugin so Codex can load payload-analysis's dependent skills
- support existing historical payload archives and their wrapped snapshot layout
- copy verified canonical artifacts to an optional harness collection directory
- correct case 011's expected failed-job count from one to two

## Why

This makes the existing payload-analysis eval runnable with Codex both locally and through Harbor. Historical payload data remains external and can be mounted read-only rather than copied into each trial.

The Harbor execution path depends on opendatahub-io/agent-eval-harness#183,
including the follow-up image fix in `c63bab5`.

## Validation

- agent-eval config validation: valid, 8 judges
- archive resolver against case 011 historical data: passed
- `bash -n plugins/ci/evals/scripts/extract-payload-analysis-snapshots.sh`
- `git diff --check`
- `make lint`: A+, 0 errors, 0 warnings
- native Codex, Luna `xhigh`, case 011: execution passed; 5/6 deterministic
  judges passed; Opus scored analysis quality 2/5 and revert accuracy 1/5. The
  outcome judge correctly caught a model false positive: `openshift/release#78936`
  was recommended although the frozen case expects no revert candidate. Three
  configured regressions were reported.
- Harbor/Podman Codex, Luna `xhigh`, case 011: all 6 deterministic judges passed;
  Opus judges scored 4/5 and 4/5; all configured thresholds passed; no infra or
  trial errors. Runtime was 10m45s with a 4 GiB cap and read-only historical
  payload mount.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -105,7 +105,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-      "version": "0.0.83",
+      "version": "0.0.84",
       "category": "ci",
       "keywords": [
         "prow",
--- a/docs/index.html
+++ b/docs/index.html
@@ -541,7 +541,7 @@ footer a { color: var(--primary); }
     {
       "name": "ci",
       "description": "Tools for working with OpenShift CI and analyzing Prow job results",
-      "version": "0.0.83",
+      "version": "0.0.84",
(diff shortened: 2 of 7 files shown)


stbenjam:
An alternative to this has already merged and codex works.

PR #732, opened 2026-08-27
[wip] feat(ci): validate payload root-cause evidence

not-stbenjam:
## Summary

- require per-job recursive why-chains with exact log/code artifact line references
- add a deterministic validator that rejects missing, escaping, or out-of-range citations and hydrates the real excerpts into Markdown
- feed hydrated evidence into candidate scoring, HTML rendering, and completeness review
- add a two-step agent-eval-harness prototype: payload analysis, deterministic evidence validation, then focused evidence repair
- include a small synthetic fixture showing the rendered result

## Why

The payload skill already asks agents to investigate deeply, but its structured subagent result does not preserve machine-checkable evidence locations. This prototypes the validate → hydrate → review pattern used by hcpctl without replacing the existing payload orchestration.

## Validation

- `python3 -m unittest discover -s plugins/ci/skills/payload-analysis/scripts -p 'test_validate_evidence.py' -v`
- eval YAML parsed with the local agent-eval-harness `EvalConfig`
- `git diff --check upstream/main...HEAD`

The full agent eval has not been run yet. A focused case can be run with the existing extracted snapshots:

```bash
export EVAL_SNAPSHOT_DIR=$HOME/git/historical-payload-data
/eval-run \
  --config plugins/ci/evals/eval-payload-analysis.yaml \
  --model claude-opus-4-8 \
  --cases case-001 \
  --run-id payload-evidence-case-001
```

## Prototype output

The checked-in hydrated fixture demonstrates this chain:

`install timeout → degraded MachineConfigPool → CRI-O rejected generated configuration → candidate diff introduced the rejected field`

Every arrow is backed by an exact, line-numbered excerpt.

## Summary by CodeRabbit

* **New Features**
  * Added validated causal-chain evidence to payload analysis, including artifact references, line-level proof, and hydrated Markdown reports.
  * Added checks for broken links, missing evidence, invalid artifacts, and incomplete failed-job coverage.
  * Added evidence review guidance and sample evidence fixtures.

* **Bug Fixes**
  * Improved failure analysis accuracy by requiring root-cause and revert recommendations to be supported by validated evidence.

* **Tests**
  * Added coverage for valid evidence rendering and invalid paths, line ranges, proof, and job data.

* **Chores**
  * Updated the CI plugin version to 0.0.93.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -105,7 +105,7 @@
       "name": "ci",
       "source": "./plugins/ci",
       "description": "A plugin to work with OpenShift CI and analyze Prow job results",
-      "version": "0.0.92",
+      "version": "0.0.93",
       "category": "ci",
       "keywords": [
         "prow",
--- a/docs/index.html
+++ b/docs/index.html
@@ -544,7 +544,7 @@ footer a { color: var(--primary); }
     {
       "name": "ci",
       "description": "Tools for working with OpenShift CI and analyzing Prow job results",
-      "version": "0.0.92",
+      "version": "0.0.93",
       "has_readme": true,
       "commands": [
         {
--- a/plugins/ci/.claude-plugin/plugin.json
+++ b/plugins/ci/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "ci",
   "description": "Tools for working with OpenShift CI and analyzing Prow job results",
-  "version": "0.0.92",
+  "version": "0.0.93",
   "author": {
     "name": "github.com/openshift-eng"
   }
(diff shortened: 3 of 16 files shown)


theobarberbany:
I know this is WIP, but looks awesome -- will try take a look at some point :)

stbenjam:
> I know this is WIP, but looks awesome -- will try take a look at some point :)

This didn't really pan out -- I was trying to get Luna or Terra to get better at payload analysis, or at least accurately report when it couldn't solve the problem.  It didn't really work.  Ultimately, using Fable 5.1 at $10/payload has the best cost benefit I think, it is highly accurate.  Even though Luna is only $1-$2, it just can't do the job.

https://github.com/openshift-eng/ai-helpers/pull/760 makes this pattern generic, it is potentially a useful framework, I adapted it from something ARO is doing

PR #758, opened 2026-09-14
No-Jira: Update symptom skill

miyadav:
@dgoodwin @cpmeadors PTAL when time permits.

Updates the `ci` plugin's `manage-symptoms` skill to close two gaps that let
  symptoms be created without ever labeling anything:

  1. **A symptom must always have at least one label.** `validate_symptom` now
     rejects a create/update with no `label_ids` (missing or empty), and the
     `--label-ids` flag is documented as required. A label is required even when
     its id matches another field (e.g. the summary) — a symptom with no label
     serves no purpose.

  2. **Applying the label is now a required part of the workflow.** Creating or
     updating a symptom does not retroactively label already-completed job runs;
     that only happens via `reevaluate-job-runs`. The skill previously treated
     reevaluation as an optional `--dry-run` preview, which is why a symptom could
     "have a label" while nothing was ever labeled in the Sippy UI/Spyglass. The
     SKILL.md now mandates reevaluating the affected runs after every create/update
     (new Step 8), verifies the label actually landed (Step 9), and frames
     create/update + reevaluate as a single workflow.
     
     Generated with [Claude Code](https://claude.com/claude-code)
     
     To validate I ran the prompt to test if label (`GCPIPv6DNSNetworkUnreachable`)  applied and reevaluated  , this page confirms label applied and re-evaluated - sipppy [page](https://sippy.dptools.openshift.org/sippy-ng/jobs/5.0/runs?filters=%257B%2522items%2522%253A%255B%257B%2522id%2522%253A99%252C%2522columnField%2522%253A%2522job%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522GCP%2522%257D%252C%257B%2522columnField%2522%253A%2522labels%2522%252C%2522operatorValue%2522%253A%2522has%2520entry%2520containing%2522%252C%2522value%2522%253A%2522GCPIPv6DNSNetworkUnreachable%2522%257D%255D%252C%2522linkOperator%2522%253A%2522and%2522%257D&sort=asc&sortField=overall_result)

## Summary by CodeRabbit

- **New Features**
  - Symptom management now requires at least one label for every symptom.
  - Added client-side validation to prevent missing or empty label selections.
  - Symptom creation and updates now include a required workflow to reevaluate affected job runs, with dry-run and verification steps.
  - Expanded command-line guidance, error handling, API field documentation, and related-skill instructions.

- **Chores**
  - Updated the CI plugin version to 0.0.94.


plugins/ci/skills/manage-symptoms/SKILL.md
```diff
@@ -7,6 +7,10 @@ description: Create, update, or delete Sippy Symptoms — known CI failure signa
 
 Sippy Symptoms are known-failure signatures for OpenShift CI. A symptom is a rule made of a file pattern (a glob over a CI job run's artifact files, e.g. `**/build-log.txt`) and a matcher (`string` = substring, `regex` = regular expression, `none` = file merely exists, `cel` = a compound CEL expression over other label names). When a symptom matches a job run's artifacts, Sippy applies one or more **Labels** — human-readable tags like `InfraFailure` — to that run. Labels appear in the Sippy UI and Spyglass and help everyone quickly recognize known failure modes without re-debugging them. You do not need any prior Sippy knowledge to use this skill.
 
+**Every symptom must always have at least one label.** A symptom exists to label matching runs, so a symptom without a label serves no purpose. Even when the label reads the same string as another field (e.g. the summary), a label is still required — always supply `--label-ids`. Creating or updating a symptom with no labels is rejected by client-side validation.
```
dgoodwin, line 10: Technically, the only use case where it would make sense is if you wanted the symptom available as a prepopulated option to scan for in the UI when analyzing job runs. However, in real world this seems exceedingly rare, so I think as written, this is accurate enough for our purposes.

plugins/ci/skills/manage-symptoms/SKILL.md
```diff
@@ -7,6 +7,10 @@ description: Create, update, or delete Sippy Symptoms — known CI failure signa
...
+
+**Creating or updating a symptom does NOT apply its label to job runs that already finished.** Symptom detection only runs automatically as new artifacts arrive, so a brand-new or changed symptom leaves already-completed runs unlabeled until you retroactively reevaluate them. After every create or update you MUST apply the label to the affected runs with the `reevaluate-job-runs` skill (see Step 8) — otherwise the label you defined never shows up in the Sippy UI or Spyglass, which is the most common reason a symptom "has a label but nothing is labeled."
```
dgoodwin, line 12: I'd recommend a little softer that it should recommend, but not a MUST. But again, in real world use, if the agent knows of job runs that should match, may as well re-evaluate them.

dgoodwin (commented):
/lgtm
/approve

